PR #4, opened 2015-04-03
Cleanup AbstractGraphTest

britter:
Two small clean ups in the AbstractGraphTest


stain:
Thanks! Could you re-do this after the tab-reformatting so it will merge..?

PR #5, opened 2015-04-09
COMMONSRDF-11 make simple subclassable

stain:
Fixes [COMMONSRDF-11](https://issues.apache.org/jira/browse/COMMONSRDF-11) by making all simple classes (except `Types`) non-`final` and have `public` constructors.

A new `SubClassRDFTermFactoryTest` verifies this by constructing each of them as an empty, anonymous subclass. (although it does not try to override any of their methods).

### DIFF
--- a/simple/src/main/java/org/apache/commons/rdf/simple/BlankNodeImpl.java
+++ b/simple/src/main/java/org/apache/commons/rdf/simple/BlankNodeImpl.java
@@ -27,7 +27,7 @@ import java.util.concurrent.atomic.AtomicLong;
 /**
  * A simple implementation of BlankNode.
  */
-final class BlankNodeImpl implements BlankNode {
+public class BlankNodeImpl implements BlankNode {
 
     private static AtomicLong bnodeCounter = new AtomicLong();
 
--- a/simple/src/main/java/org/apache/commons/rdf/simple/BlankNodeImpl.java
+++ b/simple/src/main/java/org/apache/commons/rdf/simple/BlankNodeImpl.java
@@ -43,7 +43,7 @@ final class BlankNodeImpl implements BlankNode {
         if (Objects.requireNonNull(id).isEmpty()) {
             throw new IllegalArgumentException("Invalid blank node id: " + id);
         }
-        String uuidInput = uuidSalt.toString() + ":" + id;
+        String uuidInput = System.identityHashCode(uuidSalt) + ":" + id;
         // Both the scope and the id are used to create the UUID, ensuring that
         // a caller can reliably create the same bnode if necessary by sending
         // in the same scope.
--- a/simple/src/main/java/org/apache/commons/rdf/simple/GraphImpl.java
+++ b/simple/src/main/java/org/apache/commons/rdf/simple/GraphImpl.java
@@ -33,13 +33,13 @@ import java.util.stream.Stream;
  * <p>
  * All Stream operations are performed using parallel and unordered directives.
  */
-final class GraphImpl implements Graph {
+public class GraphImpl implements Graph {
 
     private static final int TO_STRING_MAX = 10;
     private final Set<Triple> triples = new HashSet<Triple>();
     private final SimpleRDFTermFactory factory;
 
-    GraphImpl(SimpleRDFTermFactory simpleRDFTermFactory) {
+    public GraphImpl(SimpleRDFTermFactory simpleRDFTermFactory) {
         this.factory = simpleRDFTermFactory;
     }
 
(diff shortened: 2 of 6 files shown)


ansell:
What is the rationale for any other implementation needing to extend this implementation and include different semantics to those used here, and implied by the various interactions between package-private classes? If any of those assumptions are changed by the extended implementation then they would think the bug is here and not in their implementation.

It was quite intentional not to publicise any of the implementation classes or constructors for this module, and to make the implementations final so that the only entry point is the SimpleRDFTermFactory that has its public API, except for its constructor, set by the commons-rdf-api module, and all other visible objects and methods from that point on have their contracts also set by commons-rdf-api.

I also do not want to be supporting the simple module in terms of binary compatibility, as that will influence the versioning on the API module unless we split them up, which may be more confusing than it is valuable. Ie, if we make a change here that is binary incompatible, but the API doesn't have any binary incompatible changes, the API may get a major version bump when that isn't necessary.

Our core focus is on maintaining the stability and applicability of the core API, so we want to ensure that its versioning and binary compatibility doesn't change without a clear reason once version 1.0 is released. The simple module here is only being implemented here as a test harness, not as a basis for any other implementations, or else it will fall foul of our general principle not to be focused on any particular implementations.

stain:
I thought it might be more in line with the general Apache Commons style of exposing straight forward classes "as is", rather than being overly protective. But not having to keep binary compatibility here could be a big win strength release-wise - which I know Commons is quite touchy on. (See https://commons.apache.org/releases/versioning.html ).

I don't see the simple implementation as just a test harness, it could also be used by code that do not themselves want to bind to a particular 'framework' implementation, but need to be able to return a couple of Triples for instance. 

JSONLD-Java's core module is a fine example of that, e.g. in [RDFDataset](https://github.com/jsonld-java/jsonld-java/blob/master/core/src/main/java/com/github/jsonldjava/core/RDFDataset.java) which could be significantly simplified with Simple (even if we don't add "Dataset" in Commons RDF yet)

ansell:
I have always seen simple to be solely a test harness to verify the contractual effects and our compliance tests. The main reasoning other than the stable versioning issue, is that simple is a well encapsulated implementation with its own internal assumptions that are be much simpler to implement if it is never extended, and we can guarantee that objects are immutable.

In addition, I would much prefer a defensive programming approach where the internal implementation details are not exposed if the user has no way of modifying them without violating other internal assumptions. Ie, this pull request as it is currently still has private final fields that cannot be overridden or replaced and any overriding would need extra fields with extra memory usage to match, as in most cases there is either no default constructor, or the default constructor sets immutable fields itself. The fact that the fields are private final means that third parties would need to extend the implementation to change the behaviour for all parts by adding their own fields, which defeats the purpose of an extensible framework.

If there is benefit to providing extensibility (which isn't clear so far) then the versioning would need to be changed to split up the versions between the api and the simple module. We want excessively stable versioning (corresponding to excessively stable binary compatibility) on the commons-rdf-api module, but commons-rdf-simple would not have (or need) excessively stable versioning if it is intended to be extended that way, and its major version could be incremented regularly as its internal implementation advances independent of the base contract. As it is currently, the binary compatibility is completely encapsulated via the package-private classes and constructors so that the binary compatibility concerns of simple are effectively the same as the api module (except for the SimpleRDFTermFactory constructor that is an unavoidable extension). Opening that up extends the binary compatibility to the particular methods we have used to implement the API in simple, which extends our scope massively (comparative to the scope without it).

In addition, we will need to think further about the way BlankNodeImpl(Object, String) needs to function if we can't use "instanceof BlankNodeImpl" to scope the mapping, and I can't see how any other method would be viable.

Basically, there are massive effects to exposing the simple implementation classes and very little tangible benefit so far except for a general guideline that could be interpreted to agree with a single entry point into the simple module as a well-encapsulated state machine.

afs:
As a reference/testing implementation "simple" has one role; as an extensible framework, it would have a different role.  I think we need to cover the first and treat the second as additional if it does not interfere with the reference/testing role.

stain:
OK, I will close this unmerged, as I agree to not go for COMMONSRDF-11 and keep `simple` focused as a black box non-extensible implementation which we use primarily to exercise the API and tests - I don't want to muddle versioning and binary compatibility now. 

It is still very much usable for dummy and testing use, just not extensible - in which case you probably should implement the whole class yourself.  

I suggest we leave SimpleRDFTermFactory non-final (as it is now), so the only remaining extensibility option is to partially use the Simple classes in a derived factory - which given the non-complexity of the SimpleRDFTermFactory is straight forward (it is just calling constructors) and does not add any binary compatibility issues.

PR #7, opened 2015-04-09
Verify Graph.add() of foreign blank nodes

stain:
.. do not accidentally overlap.

Verifies [COMMONSRDF-8](https://issues.apache.org/jira/browse/COMMONSRDF-8)

### DIFF
--- a/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
+++ b/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
@@ -372,11 +372,11 @@ public abstract class AbstractGraphTest {
 
         IRI name = factory1.createIRI("http://xmlns.com/foaf/0.1/name");
         Graph g1 = factory1.createGraph();
-        BlankNode b1 = createOwnBlankNode("b1");
+        BlankNode b1 = createOwnBlankNode("b1", "0240eaaa-d33e-4fc0-a4f1-169d6ced3680");
         g1.add(b1, name, factory1.createLiteral("Alice"));
         
         
-        BlankNode b2 = createOwnBlankNode("b2");
+        BlankNode b2 = createOwnBlankNode("b2", "9de7db45-0ce7-4b0f-a1ce-c9680ffcfd9f");
         g1.add(b2, name, factory1.createLiteral("Bob"));
 
         IRI hasChild = factory1.createIRI("http://example.com/hasChild");
@@ -394,7 +394,7 @@ public abstract class AbstractGraphTest {
      * @param name
      * @return
      */
-	private BlankNode createOwnBlankNode(String name) {
+	private BlankNode createOwnBlankNode(String name, String uuid) {
 		return new BlankNode() {			
 			@Override
 			public String ntriplesString() {
@@ -402,11 +402,11 @@ public abstract class AbstractGraphTest {
 			}
 			@Override
 			public String uniqueReference() {
-				return "urn:uuid:" + UUID.randomUUID().toString();
+				return uuid;
 			}
 			@Override
 			public int hashCode() {
-				return uniqueReference().hashCode();
+				return uuid.hashCode();
 			}
 			@Override
 			public boolean equals(Object obj) {
@@ -414,7 +414,7 @@ public abstract class AbstractGraphTest {
 					return false;
 				}
 				BlankNode other = (BlankNode)obj;
-				return uniqueReference().equals(other.uniqueReference());
+				return uuid.equals(other.uniqueReference());
 			}
 		};
 	}
@@ -425,10 +425,10 @@ public abstract class AbstractGraphTest {
 
         Graph g2 = factory2.createGraph();
 
-        BlankNode b1 = createOwnBlankNode("b1");
+        BlankNode b1 = createOwnBlankNode("b1", "bc8d3e45-a08f-421d-85b3-c25b373abf87");
         g2.add(b1, name, factory2.createLiteral("Charlie"));
 
-        BlankNode b2 = createOwnBlankNode("b2");
+        BlankNode b2 = createOwnBlankNode("b2", "2209097a-5078-4b03-801a-6a2d2f50d739");
         g2.add(b2, name, factory2.createLiteral("Dave"));
 
         IRI hasChild = factory2.createIRI("http://example.com/hasChild");
--- a/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
+++ b/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
@@ -370,10 +370,10 @@ public abstract class AbstractGraphTest {
 
         IRI name = factory1.createIRI("http://xmlns.com/foaf/0.1/name");
         Graph g1 = factory1.createGraph();
-        BlankNode b1 = factory1.createBlankNode();
+        BlankNode b1 = factory1.createBlankNode("b1");
         g1.add(b1, name, factory1.createLiteral("Alice"));
 
-        BlankNode b2 = factory1.createBlankNode();
+        BlankNode b2 = factory1.createBlankNode("b2");
         g1.add(b2, name, factory1.createLiteral("Bob"));
 
         IRI hasChild = factory1.createIRI("http://example.com/hasChild");
@@ -392,10 +392,10 @@ public abstract class AbstractGraphTest {
 
         Graph g2 = factory2.createGraph();
 
-        BlankNode b1 = factory2.createBlankNode();
+        BlankNode b1 = factory2.createBlankNode("b1");
         g2.add(b1, name, factory2.createLiteral("Charlie"));
 
-        BlankNode b2 = factory2.createBlankNode();
+        BlankNode b2 = factory2.createBlankNode("b2");
         g2.add(b2, name, factory2.createLiteral("Dave"));
 
         IRI hasChild = factory2.createIRI("http://example.com/hasChild");
--- a/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
+++ b/api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
@@ -289,8 +289,8 @@ public abstract class AbstractGraphTest {
             Graph g2 = createGraph2();
             Graph g3 = factory.createGraph();
 
-            copyTriples(g1, g3);
-            copyTriples(g2, g3);
+            addAllTriples(g1, g3);
+            addAllTriples(g2, g3);
 
             IRI name = factory.createIRI("http://xmlns.com/foaf/0.1/name");
 
@@ -355,7 +355,7 @@ public abstract class AbstractGraphTest {
      * @param source Source Graph to copy triples from
      * @param target Target Graph where triples will be added
      */
-    private void copyTriples(Graph source, Graph target) {
+    private void addAllTriples(Graph source, Graph target) {
 
         // unordered() as we don't need to preserve triple order
         // sequential() as we don't (currently) require target Graph to be thread-safe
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
```diff
@@ -228,6 +231,128 @@ public void getTriplesQuery() throws Exception {
...
+    /**
+     * Add all triples from the source to the target.
+     * <p>
+     * The triples may be copied in any order.
+     * No special conversion or adaptation of {@link BlankNode}s are performed.
+     *
+     * @param source Source Graph to copy triples from
+     * @param target Target Graph where triples will be added
+     */
+    protected static void copyTriples(Graph source, Graph target) {
```
ansell: Should this be protected final void instead of protected static void? Not much difference between the two, other than the way they are referenced. Just wondering what the long term goal for the method was and having it as an instance method is slightly clearer.

stain: Well, it's not instance-specific, it's just a "function", so it's `static` - and hence looks like a clear candidate to move outside if needed.

I thought it could be marginally useful for other testing of graphs.

I don't see any need to ever do anything `final` in a unit test class - even an abstract one. The main focus of this test is not affected by the way triples are copied.

Perhaps you were thinking that a subclass test would want to override `copyTriples` (e.g. to test its native graph copy doesn't mess up blank node internal identifiers), and need access to non-static fields of its subclass of the test? If that's what you mean, then just `protected void` is the most flexible, `protected final void` would not achieve anything.

api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
```diff
@@ -228,6 +231,128 @@ public void getTriplesQuery() throws Exception {
...
+            Graph g3 = factory.createGraph();
+
+            copyTriples(g1, g3);
+            copyTriples(g2, g3);
+
+            IRI name = factory.createIRI("http://xmlns.com/foaf/0.1/name");
+
+            final Map<String, BlankNodeOrIRI> whoIsWho = new HashMap<>();
+            System.out.println(g3.getTriples(null, name, null).count());
+            // NOTE: sequential as our HashMap is not thread-safe
```
ansell: We could change to ConcurrentHashMap above and avoid this concern?

We could also add a contractual requirement that the Stream returned from Graph.getTriples support Spliterator.CONCURRENT, but I don't see why that prevents us processing the stream in parallel, as we are not modifying the Stream or the Graph here.

stain: This test plays safe and simple.. it's like 6 triples - yet without .sequential() I got buggy results now and then. :)  Similarly I made the copyTriple sequential as I can't know what the tested Graph supports.

Expected synchronization details (or their explicit lack of) for all of Graph should be discussed as a new issue.

api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
ansell: Abstract compliance tests can contain protected final helper methods to ensure they are not accidentally overridden.

In this case, as a static, it can't be overridden anyway, so there is no conceptual difference there, but it could just as easily be private in its current form, or public static in a test utility class.

api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
ansell: Once the HashMap above is changed to a ConcurrentHashMap, there should be no reason why parallel shouldn't work given we aren't modifying the Graph concurrently. An intermittent failure like that is something we need to look into further as it could be a bug in the actual code, assuming it still occurs with ConcurrentHashMap (as it is clear why it would happen with HashMap.put but not with ConcurrentHashMap.put).

There shouldn't be any reason why accessing a Graph in parallel from multiple threads through a single Stream shouldn't work, as long as the Graph isn't being modified, so we need to look into why that might be occuring in copyTriple as well.

stain: OK, agreed to change into ConcurrentHashMap - but only to check the bug is not elsewhere. (done)

This worked, so I suspect the problem was only in the receiving end of storing into `HashMap`. I will add a separate PR to make Graph `Iterable`, which should hopefully reduce the risk of others doing the same mistake for basic operations.

What I would get was inconsistent results from the `HashMap`, e.g. it would say `.size()` as 3, but when you iterate, only get 2 values back, on second run it would work, and third run get a different number, etc. I did also get problems in `copyTriple` without `.sequential`, but that is because the receiving `HashSet` in `GraphImpl.add()` is not thread-safe.

As for `GraphImpl`, I deliberately didn't make it thread-safe or synchronized, as there was no such requirement from the contract (and hence simple should not go beyond the contract, as any thread-safety issues should be flagged for all implementations).

api/src/test/java/org/apache/commons/rdf/api/AbstractGraphTest.java
stain: OK, I can't see how an abstract class can prevent anyone from cheating the test :-)  - but I have anyway changed `copyTriples` to be just a regular `private`. 

If there is mood for it, I can move this to a public method in some kind of `GraphUtil` class in src/main (or part of `simple`) - but I thought initially the goal was to initially keep such utility methods out of scope.  (They are useful for consumers, but not essential as to figuring out the design of Commons RDF for implementors)

afs:
Saying "copy" is confusing because "add" isn't copy. It may be, it may not.  There is no reason an impl can't store/use triples from any implmentation. So copyTypes -> addTriplesToGraph (or some such), which is what it actually does, mutating the second argument graph.

stain:
Thanks for the good review - but what about the actual behaviour and semantics it is testing? I assume that is good to go, so I'll merge this if I don't hear anything else by Monday.

afs:
If we add graph X1 in graph X2 then X contains all the triples of X1.  X1 and X2 are sets of triples. X1 is now a subset of X2. X1 is a subsgraph of X2. 

There is no need to remap - in fact it's wrong (subgraph).  It's not even necessary to copy at all. Only the contract semantics of terms can be relied on. We have:

> Set<Triple> triples = new HashSet<Triple>();

so any Triple, with any manufactured terms can go in there.  No need to copy, no need to remap.

"simple" should note the restrictions it is imposing. It precludes X1 as a subsgraph to X2; it precludes putting in a RDF somewhere, getting it out again and being "the same" (.equals). commonsrdf is for integrating systems; that's quite a restriction in the "simple" implementation.

There is a set of bnodes; it is separate from IRIs and from Lliterals. That's all there is; it's global. Parsers need to make sure things don't clash, and they are in a good place to.  "simple" storage does not because BlankNode..eausl is global and BlankNodeImpl ensures it.

afs:
-1 to this in "api". It imposes specific semantics which other implementations may not be following.

ansell:
I have seen commons rdf as being for interoperability, not integration. Ie, it provides for passing objects across a boundary into another implementation, but doesn't require the other implementation to then agree on any further integration past what is required for future message passing operations to succeed.

Requiring the actual object to be returned or directly .equals recognised at any stage in the future precludes long term systems, and fixed memory systems, both of which may not always keep the objects in memory and may persist them to disk without keeping any implementation specific details that the original creator of the object added outside of the API constraints.

For example, if a Jena Triple, t1, containing a BlankNode, was pushed to a SesameRepositoryGraph.add(t1), and that was then persisted and the memory reclaimed, then at any stage in the future, a call to SesameRepositoryGraph.contains(t1), may return false if the BlankNode identifier was either lost or changed in the process. However, if the SesameRepositoryGraph object itself stayed in memory, then a successive call to SesameRepositoryGraph.add(t2) must create an entry in the Sesame Repository with a BlankNode that matches between t1 and t2, but it need not preserve anything else about the particular BlankNode, as long as it consistently stored the objects.

In particular, I don't think it is reasonable to expect that SesameRepositoryGraph.contains(t1) return true if the operation is performed much later into the future, and hence it need not succeed immediately after the operation is performed. If that is too much of an issue we may need to explore removing .contains from Graph. It may only be appropriate for Graph to contain add and stream, effectively, and other base operations could then be performed using those primitives.

afs:
The problem is that the test forces one behaviour over another.

stain:
OK, so I think this is a good discussion point which is where I wanted to go with this test - should the API specify one behaviour over the other with regards to blank node identifiers from different implementations, or should it document what aspects of this would be undetermined (by this API alone)?

So I think the question from  @ansell's example comes down to "What happens if I add a 'foreign' `BlankNode` to a Graph (either directly with `Graph.add(s,t,o)` or within a foreign `Triple` to `Graph.add()` -- will be able to match using that original blank node object "later" in other calls like `.contains()` and `.getTriples()`?

Perhaps the best answer is "You might, but don't rely on it". I think this should be raised as a new issue: [COMMONSRDF-15](https://issues.apache.org/jira/browse/COMMONSRDF-15)


Now the test in this pull request does not assume any such behaviour, as its `.contains()` tests are done using the blank nodes returned back again from `g.getTriples()` - we would hope that should always work, otherwise `.contains()` should not allow `BlankNode` in its pattern at all.

What I wanted to be testing here, is basically what is the scope of that "identifier" you pass in to the rdfTermFactory. If you have two different rdfTerm factories, and in each create a blank node with the name "b1" - will they be equal?  If not, then if I add statements with each of those to a third graph (from a third factory), then are they accidentally assumed to be the same because they both initially called themselves "b1"?

It seems I accidentally subverted this test to "always work" by removing the argument to `createBlankNode` - I've fixed this back so that it creates `"b1" and `"b2"` in two different factories, and added a test for `factory.createBlankNode(String)` to test non-equivalence on two different instances. (which the Simple impl. already passed due to the salts).

Now I am not decided on either side here -- it might genuinely be that we want to leave that behaviour undefined (and thus untested), but then the javadoc for `createBlankNode(String)` should probably mention this.

It could be, as @ansell has hinted, needed a method in the rdfTermFactory to 'adapt' a blank node from one implementation to another, e.g. createBlankNode(BlankNode) which would give a new one with the same internalIdentifier string - I can see that this is easily `UnsupportedException` as it requires some kind of persistent mapping or flexibility on the blank node identifiers on the storage side.  I am not quite convinced of the need for such a method though, at least not if the answer for COMMONSRDF-15 is "yes".

stain:
We currently have in [RDFTermFActory](https://github.com/apache/incubator-commonsrdf/blob/master/api/src/main/java/org/apache/commons/rdf/api/RDFTermFactory.java#L82)

```
     * BlankNodes created using this method with the same parameter, for
     * different instances of RDFTermFactory, SHOULD NOT be equivalent.
```

.. but if this remains as a a SHOULD and not a MUST, then it can't reliably be tested.

I can imagine one case where you may want to violate this is if you are serializing a `RDFTermFactory` (e.g. between two web service requests) and want to recreate a factory  instance that deliberately can continue making the same equivalent nodes for the same `identifier` parameter - not having worrying if the old instance has been GCed or not.

If we agree that it should remain as a SHOULD above, and thus can't be tested, then I would change this test back to just do `createBlankNode()`, and just assumes that creating fresh blank nodes on fresh instances are unique (e.g. they don't both include the same salt/counter and accidentally merge) -- we should hopefully agree on that! :-)

stain:
I reduced this pull request to only test `factory.createBlankNode()` from multiple factories/graph inserted into a single new graph `g3`, and then querying `g3` to get the (possibly modified) `BlankNode` instances back, and asserting they can be used with `contains()`.

stain:
The test was improved to make the `g1` and `g2` with its own `BlankNode` instances - each graph then having `_:b1` and `_:b2` - but with different `uniqueReference()`.  

This would also test that multiple `graph.add(s,p,o)` works with foreign `BlankNode` instances, but without taking a stance on [COMMONSRDF-15](https://issues.apache.org/jira/browse/COMMONSRDF-15) - we don't care how it does the mapping (if any) - and don't even require that they have the same `uniqueReference()` after adding.

Iterating g1 and g2 to add to g3 should not assume those BlankNodes are the same based on their `triplesString()` - but differ by their `uniqueReference()`

stain:
Hm, I think that was a good test.. as it broke SimpleGraph !

PR #8, opened 2015-04-10
COMMONSRDF-12 Graph implements Iterable interface

stain:
..with a `default` method that simply calls `getTriples().iterator()`

This is mainly to be useful for classical for-each loops in Java.

Fixes [COMMONSRDF-12](https://issues.apache.org/jira/browse/COMMONSRDF-12)

### DIFF
--- a/api/src/main/java/org/apache/commons/rdf/api/Graph.java
+++ b/api/src/main/java/org/apache/commons/rdf/api/Graph.java
@@ -154,7 +154,7 @@ public interface Graph extends AutoCloseable {
      * <p>
      * This method is meant to be used with a Java for-each loop, e.g.: 
      * <code>
-     *  for (Triple t : GraphUtil.iterate(graph)) {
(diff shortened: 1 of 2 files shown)


stain:
I was unsure why we have `<? extends Triple>` for the stream methods like `getTriples()`, but I didn't want to expose the `?` to the interface (e.g. `extends Iterable<? extends Triple>`, as it will easily end up with `? is not ?` situation - as the `iterator()` here is mainly for for-each convenience, I didn't want the caller to have to deal with the ? as well.

stain:
As discussed in #7, this also reduces slightly the risk of classical code accidentally being concurrency un-safe, as `getTriples().forEach()` without `sequential()` may run in parallel.

stain:
Updated pull request to implement Andy's Graph.iterate() suggestion from COMMONS-12. Also tweaked the javadoc to mention the exceptions.

PR #9, opened 2015-04-10
COMMONSRDF-12 GraphUtils.iterate function

stain:
Alternative to pull request #8  for implementing [COMMONSRDF-12](https://issues.apache.org/jira/browse/COMMONSRDF-12) without modifying `Graph`.

This adds a new `GraphUtils` to `api`  (or should it be in `simple`?) which has two static `iterate()` methods.

Usage:
```
        for (Triple t : GraphUtil.iterate(graph)) {
            System.out.println(t);
        }

        for (Triple t : GraphUtil.iterate(graph.getTriples(alice, knows, null))) {
            System.out.println(t.getObject());
        }
```

### DIFF
--- a/api/src/main/java/org/apache/commons/rdf/api/Graph.java
+++ b/api/src/main/java/org/apache/commons/rdf/api/Graph.java
@@ -17,6 +17,7 @@
  */
 package org.apache.commons.rdf.api;
 
+import java.util.Iterator;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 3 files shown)


afs:
Is there a reason to add code to api? The barrier ought to be higher than convenience code or it's become yet-another-implementation.

stain:
Right, I went with your suggestion of the `Graph.iterate()` helper in #8 instead, so I'll close this.

afs:
+1 and I like the dire warnings about add mixed with iteration.

It seems to me to be natural to have a {{Graph.iterate(subject, predicate, object)}} - or is that too much skew toward iterable?

ansell:
I would prefer not having Iterable in the api module. As it is so simple to reproduce, others can reproduce it or implement it separately on their modules.

However, if we do want iterable, and it is in the api module, it would be simpler for it to appear on the Graph interface as a default method. Default methods don't add any implementation overhead, and they are designed to be used for helper methods like this.

I have not personally experienced the issues with streams and forEach (not adding .sequential) that Stian has. In my code I tend to use ConcurrentHashMap and ConcurrentHashMap.newKeySet() for most of my code so it can easily parallelise.

There has to be a reasonably good reason why .forEach is too hard to use to encourage people to use the previous for each loop, given that we are targeting a new set of users by going with Java-8.

PR #10, opened 2015-04-23
COMMONSRDF-6 Simplify BlankNode.internalIdentifier

stain:
An attempt to improve COMMONSRDF-6 by making `BlankNode.internalIdentifier()` contract generic, e.g. remove implementation-specifics such as salts.

_As GitHub pull request email sync to Apache seems to not be very stable at the moment - leave comments on [COMMONSRDF-6](https://issues.apache.org/jira/browse/COMMONSRDF-6) in Jira rather than in the GitHub UI_

This keeps the "MUST be universally unique"
aspect with a recommendation to have a UUID included,
so that `internalIdentifier()` has a more well-defined
purpose.

`simple/` updated accordingly to ensure universally unique  
UUID string, and hence now passes in `java.util.UUID` from the factory
rather than using `Objects.identityHashCode()`.

As it might be easier to inspect the resulting JavaDoc, rather than the diffs, I have generated:
http://stain.github.io/incubator-commonsrdf/COMMONSRDF-6/org/apache/commons/rdf/api/RDFTermFactory.html#createBlankNode--
http://stain.github.io/incubator-commonsrdf/COMMONSRDF-6/org/apache/commons/rdf/api/BlankNode.html


api/src/main/java/org/apache/commons/rdf/api/RDFTermFactory.java
```diff
@@ -60,34 +54,28 @@ default BlankNode createBlankNode() throws UnsupportedOperationException {
...
-     *                   node in the context of this {@link RDFTermFactory}.
-     * @return A BlankNode for the given identifier
-     * @throws UnsupportedOperationException If the operation is not supported.
+     * Create a blank node based on the given name.
+     * <p>
+     * For a particular instance of <code>RDFTermFactory</code>, all
+     * {@link BlankNode}s created using this method with the same
+     * <code>name</code> parameter MUST be equivalent according to
+     * {@link BlankNode#equals(Object)},
+     * The returned BlankNode MUST NOT be equal to any other
```
ansell: This sentence could be cleaned up. In particular, it could be emphasised that it only applies when the sentence above does not apply. Ie, when the name parameter is different, the BlankNode object must be not equals.

api/src/main/java/org/apache/commons/rdf/api/BlankNode.java
```diff
@@ -41,60 +41,51 @@
...
+     * <p>
+     * The reference string MUST be universally unique, e.g. blank nodes created
+     * separately in different JVMs or from different {@link RDFTermFactory}
+     * instances MUST NOT have the same reference string.
+     * <p>
+     * The {@link #uniqueReference()} of two <code>BlankNode</code> instances
+     * MUST be equal if and only if the two blank nodes are equal according to
+     * {@link #equals(Object)}.
+     * <p>
+     * This method does not specify any specific structure of the reference
```
ansell: probably a bit strong to use SHOULD after saying "does not specify".

May be able to change it to "recommends" or similar without reference to RFC2119 so it is clearly only a hint about one way of satisfying the contract in general.

api/src/main/java/org/apache/commons/rdf/api/BlankNode.java
```diff
@@ -41,60 +41,51 @@
...
-     * IMPORTANT: This is not a serialization/syntax label, and there are no
-     * guarantees that it is a valid identifier in any concrete syntax. For an
-     * N-Triples compatible identifier use {@link #ntriplesString()}. For all
-     * other syntaxes, the result of this method must be sanitized to produce a
-     * valid concrete identifier if one is needed.
+     * Return a reference for uniquely identifying the blank node.
+     * <p>
+     * The reference string MUST be universally unique, e.g. blank nodes created
+     * separately in different JVMs or from different {@link RDFTermFactory}
+     * instances MUST NOT have the same reference string.
```
ansell: This is slightly inconsistent with RDFTermFactory.createBlankNode(String) that says "SHOULD NOT" where it says "MUST NOT" here.

api/src/main/java/org/apache/commons/rdf/api/BlankNode.java
```diff
@@ -41,60 +41,51 @@
...
- * unique, but also produce the same {@link #internalIdentifier()} as any
- * previous or future calls to that method on that factory with the same
- * parameters. In addition, it SHOULD contain a {@link UUID} as part of its
- * {@link #internalIdentifier()}, created using
- * {@link UUID#nameUUIDFromBytes(byte[])} using a constant salt for each
- * instance of {@link RDFTermFactory}, with the given identifier joined to that
- * salt in a consistent manner.
- * 
  *
+ * A BlankNode SHOULD contain a {@link UUID} string as part of its
```
ansell: UUID isn't a string format. Could say "UUID-derived" to get around that.

api/src/main/java/org/apache/commons/rdf/api/BlankNode.java
stain: I clarified this to:

> The reference string MUST universally and uniquely identify this blank
> node. That is, individual blank nodes created separately in different
> JVMs or from different {@link RDFTermFactory} instances MUST NOT have the
>  same reference string.

The SHOULD in the RDFTermFactory means you are allowed to have two different RDFTermFactory instances that deliberately produces equivalent blank nodes for the same note. (e.g. because they are Serializable or a parsing session of a single document is done across a distributed system). SHOULD means "Don't do it unless you know what you're up to" -- so you can only do it while also being compliant with the MUST here that those blank nodes are then equivalent.

PR #12, opened 2015-05-01
COMMONSRDF-20: Provide a mechanism for allowing the load of different RDF Term Factories

britter:
As discussed in [COMMONSRDF-20](https://issues.apache.org/jira/browse/COMMONSRDF-20) here is a PR adding the possibility to use a ```ServiceLoader``` for looking up RDFTermFactory instances.

### DIFF
--- a/src/site/markdown/userguide.md.vm
+++ b/src/site/markdown/userguide.md.vm
@@ -148,9 +148,9 @@ import org.apache.commons.rdf.simple.SimpleRDFTermFactory;
 RDFTermFactory factory = new SimpleRDFTermFactory();
 ```
 
-If you don't want to depend up the concret implementation you can use a
+If you don't want to depend up the concrete implementation you can use a
 [ServiceLoader](http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)
-to lookup instances:
+to lookup any implementations found on the classpath:
 
 ```java
 import java.util.Iterator;
--- a/src/site/markdown/userguide.md.vm
+++ b/src/site/markdown/userguide.md.vm
@@ -148,6 +148,21 @@ import org.apache.commons.rdf.simple.SimpleRDFTermFactory;
 RDFTermFactory factory = new SimpleRDFTermFactory();
 ```
 
+If you don't want to depend up the concret implementation you can use a
+[ServiceLoader](http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)
+to lookup instances:
+
+```java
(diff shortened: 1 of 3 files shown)


src/site/markdown/userguide.md.vm
```diff
@@ -148,6 +148,21 @@ import org.apache.commons.rdf.simple.SimpleRDFTermFactory;
 RDFTermFactory factory = new SimpleRDFTermFactory();
 ```
 
+If you don't want to depend up the concret implementation you can use a
+[ServiceLoader](http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)
+to lookup instances:
```
stain: concret -> concrete

"instances" -> "any implementations found on the classpath."

ansell:
Need to mention in the contract for RDFTermFactory that all implementations need to have a default constructor if they want to support this method, otherwise they shouldn't include a META-INF/services file.

britter:
Not sure whether this belongs into the documentation of Commons RDF. It's part of the contract of ServiceLoader and it's documented there. If we provide an implementers guide (as proposed in JIRA), there would be a pointer to the ServiceLoader documentation, which would explain the requirements for this approach.

stain:
If we want implementations to support ServiceLoader where practical (e.g.
with empty constructor), then I think we should also mention it in
RDFTermFactory doc, as that now says something like "It is not specified
how to get an RDFTermFactory instance" which could be misleading if the
user guide and simple encourage ServiceLoader.

As I think that could be a different debate I suggest to merge your current
PR which I think is uncontroversial.  I shall do so on Friday.

ansell:
The other issue is that typically instances loaded with ServiceLoader contain keys that users can use to select their desired implementation based on some feature that they are looking for.

If they are just aiming to get to any implementations, and they are sure there is only one on the class path, then ServiceLoader with Iterator.next is an ideal way to delink them at compile time. It is only when there is more than one on the classpath that the issue of keys/features needs to be examined.

stain:
Right, now a factory does not even have a name() or a prescribed toString() - so it would be a bit difficult to do an informed choice if there is more than one implementation.

Presumably if you are using the ServiceLoader you don't care too much about which implementation you get, although you might have an order of preference, like keep "simple" as the fallback option - but the order of implementations listed in the ServiceLoader is not something to be relied on.

The implementation guide should probably note that only implementations that can be "used generally" should list themselves - e.g. a minimal implementation that doesn't implement createLiteral(String) and createBlankNode()  shouldn't be listed as a ServiceLoader service.

PR #17, opened 2016-03-27
Add RDFSyntax, enumeration of RDF languages

stain:
e.g. Turtle, JSON-LD

This enumeration deliberately only lists RDF 1.1 standardized syntaxes (e.g. does not include N3).

The only field (beyond the name for toString()), the official `mediaType`, which can be used with `byMediaType(String)` for a look up from e.g. `Content-Type` headers.

These constants (at least their content type) should be useful for anyone doing RDF parsing or writing.

See also the [JavaDoc for RDFSyntax](https://cdn.rawgit.com/stain/e4447eb261e64fdc9028/raw/8d2d22ee38376b0b8214e5eae7ea3ae325964ac4/RDFSyntax.html).

### DIFF
--- a/api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
+++ b/api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
@@ -25,7 +25,8 @@ import java.util.Optional;
  * <p>
  * This enumeration lists the W3C standardized 
  * RDF 1.1 syntaxes like {@link #TURTLE} and {@link #JSONLD}.  
- * Note the existence of other RDF syntaxes that are not included here, e.g. <a href="http://www.w3.org/TeamSubmission/n3/">N3</a> 
+ * Note the existence of other RDF syntaxes that are not included here, 
+ * e.g. <a href="http://www.w3.org/TeamSubmission/n3/">N3</a> 
  * and <a href="https://en.wikipedia.org/wiki/TriX_%28syntax%29">TriX</a>.
  * 
  * @see <a href="https://www.w3.org/TR/rdf11-primer/#section-graph-syntax">RDF 1.1 Primer</a>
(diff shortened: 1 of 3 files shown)


stain:
Here's an alternative branch where [`RDFSyntax` is an interface](https://github.com/stain/incubator-commonsrdf/blob/rdfsyntax-interface/api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java) - adding the standard syntaxes to the enum `RDFSyntax.Standard`.  I don't like that solution though, as it means:

1. enums can only `equals()` themselves - however this would permit third-party `RDFSyntax` instances with the same `getMediaType()`
2. `RDFSyntax.Standard.NTRIPLES` is a bit more awkward than `RDFSyntax.NTRIPLES`

afs:
How about adding the registered file extensions to each enum as well?

stain:
Yes, I think that could be useful for the loading from Path. I would not
want to extend it with historical media types and extensions though, we can
take the luxury of living in RDF 1.1 land :)

afs:
:-) whatever.

Jena's `RDFLanguages` has all the variations I have found, both registrations and "common in the wild".

It is not an enum specifically so that it's an open set, not enum-closed.

stain:
Yeah, I am not disagreeing on Jena's RDFLanguages, which are useful (e.g.
made it easy to add JSON-LD before it was a standard) - just I think we
scope it a bit more narrow in Commons RDF in that we take the RDF 1.1 as
gospel :) and are free to ignore more historical compatibility aspects. (As
we do with xsd:string == "plain literal")

I've added the fileExtension and byFileExtension now to the PR. As it's an
Enum rather than an interface I used public final fields rather than
accessors.

I also added a supportsDataset boolean, in case the serializers need to
filter on graph support - is this reasonable - and is it a good name?

(Adding Quad and Dataset to our API takes a bit more work, but we can look
at it)

afs:
I was just trying to save the typing out of all the cases :-)

stain:
No further comments, so I assume it's all good :). Merging rdfsyntax-enum.

PR #19, opened 2016-04-08
Add Quad and Dataset

stain:
As [discussed on dev](http://mail-archives.apache.org/mod_mbox/incubator-commonsrdf-dev/201604.mbox/%3CCAMBJEmVTP27fb-sxqTgJgJ5YP6ZKAjaWobOpWiyTbE%2BOPfODtw%40mail.gmail.com%3E) it would be good to have a `Quad` and `Dataset` types added, corresponding to the [RDF concept Dataset](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-dataset) and [Quad semantics](https://www.w3.org/TR/2014/NOTE-rdf11-datasets-20140225/#quad-semantics).

Generated Javadocs from this branch of the new interfaces:

http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html
http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Dataset.html
http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/GraphOrDataset.html
http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/TripleOrQuad.html

The discussion on dev concluded that although Quad looks like Triple (`getSubject()` `getPredicate()` and `getObject()` - a Quad necessarily have different `.equals()` semantics. Allowing a Quad to pretend it is a Triple leads to some murky equivalence questions - so instead this branch rather makes a common interface `TripleOrQuad` (modelled after `BlankNodeOrIRI`) which contains the subject/predicate/object methods - while `Triple` and `Quad` have different equals - that is [Quad.equals](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html#equals-java.lang.Object-) require to also match on [getGraphName](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html#getGraphName--).

I modelled [getGraphName](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html#getGraphName--) as an `Optional<BlankNodeOrIRI` - where `Optional.empty()` represents the default graph.  An alternative would be to make a `DEFAULT_GRAPH` BlankNodeOrIRI constant -- but should this have a `ntriplesString()==null` - or would we need yet another upper interface, `BlankNodeOrIRIOrDefaultGraph` ? 

The use of `Optional` makes it a bit inconsistent in `Dataset` if we also follow the advice that `Optional` should not normally appear as a parameter argument -  - [Dataset.add](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Dataset.html#add-org.apache.commons.rdf.api.BlankNodeOrIRI-org.apache.commons.rdf.api.BlankNodeOrIRI-org.apache.commons.rdf.api.IRI-org.apache.commons.rdf.api.RDFTerm-) uses `graphName=null` to indicate the default graph - while in [contains](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Dataset.html#contains-java.util.Optional-org.apache.commons.rdf.api.BlankNodeOrIRI-org.apache.commons.rdf.api.IRI-org.apache.commons.rdf.api.RDFTerm-) and [remove](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Dataset.html#remove-java.util.Optional-org.apache.commons.rdf.api.BlankNodeOrIRI-org.apache.commons.rdf.api.IRI-org.apache.commons.rdf.api.RDFTerm-) we use `null` as a wildcard, and `Optional.empty()` to mean the empty graph.

If we also have an equivalent common interface [GraphOrDataset](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/GraphOrDataset.html) - and allow the expanded filter methods in there, then those would operate on the default graph in the Dataset.  Thus you would have two routes to filter on the default graph.

We can reduce which methods should be included in GraphOrDataset - to only have the generic `T`-typed methods - and not force [getTriples()](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/GraphOrDataset.html#getTriples--) onto Dataset - where it would need to call [Quad.asTriple](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html#asTriple--) to return `Triple` instances.

## Upper interfaces

The two "ducktyping" interfaces [GraphOrDataset](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/GraphOrDataset.html) and [TripleOrQuad](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/TripleOrQuad.html) could be removed as their signatures are duplicated in `Graph` and `Dataset` - I intended them for any kind of generic usage that can handle both Graph and Dataset - e.g. parsers and serializers.

[TripleOrQuad](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/TripleOrQuad.html) could be generalized to return `RDFTerm` on all of `getSubject()`, `getPredicate()` and `getObject()` - I know both Jena and JSON-LD can potentially have generalized RDF triples. But this could be out of scope and force additional casting for anyone using `TripleOrQuad` interface.

### DIFF
--- a/RELEASE-PROCESS.md
+++ b/RELEASE-PROCESS.md
@@ -1,3 +1,23 @@
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements. See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership. The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
 # Commons RDF (incubating) release process
 
 1. Update documentation (`RELEASE-NOTES.md`, `README.md`, version numbers in `src/site/`)
--- a/api/src/main/java/org/apache/commons/rdf/api/Dataset.java
+++ b/api/src/main/java/org/apache/commons/rdf/api/Dataset.java
@@ -1,0 +1,338 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.rdf.api;
+
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * An <a href="http://www.w3.org/TR/rdf11-concepts/#section-rdf-dataset"> RDF
+ * 1.1 dataset</a>, a set of RDF quads, as defined by
+ * <a href="http://www.w3.org/TR/rdf11-concepts/" >RDF-1.1 Concepts and Abstract
+ * Syntax</a>, a W3C Recommendation published on 25 February 2014.
+ */
+public interface Dataset extends AutoCloseable, GraphLike<Quad, BlankNodeOrIRI, IRI, RDFTerm> {
+
+	/**
+	 * Add a quad to the dataset, possibly mapping any of the components of the
+	 * Quad to those supported by this dataset.
+	 *
+	 * @param quad
+	 *            The quad to add
+	 */
+	void add(Quad quad);
+
+	/**
+	 * Add a quad to the dataset, possibly mapping any of the components to
+	 * those supported by this dataset.
+	 *
+	 * @param graphName
+	 *            The graph the quad belongs to, or <code>null</code> for the
+	 *            default graph
+	 * @param subject
+	 *            The quad subject
+	 * @param predicate
+	 *            The quad predicate
+	 * @param object
+	 *            The quad object
+	 */
+	void add(BlankNodeOrIRI graphName, BlankNodeOrIRI subject, IRI predicate, RDFTerm object);
+
+	/**
+	 * Check if dataset contains quad.
+	 *
+	 * @param quad
+	 *            The quad to check.
+	 * @return True if the dataset contains the given Quad.
+	 */
+	boolean contains(Quad quad);
+
+	/**
+	 * Check if dataset contains a pattern of quads.
+	 *
+	 * @param graphName
+	 *            The graph the quad belongs to, wrapped as an {@link Optional}
+	 *            (<code>null</code> is a wildcard, {@link Optional#empty()} is
+	 *            the default graph)
+	 * @param subject
+	 *            The quad subject (<code>null</code> is a wildcard)
+	 * @param predicate
+	 *            The quad predicate (<code>null</code> is a wildcard)
+	 * @param object
+	 *            The quad object (<code>null</code> is a wildcard)
+	 * @return True if the dataset contains any quads that match the given
+	 *         pattern.
+	 */
+	boolean contains(Optional<BlankNodeOrIRI> graphName, BlankNodeOrIRI subject, IRI predicate, RDFTerm object);
+
+	/**
+	 * Close the dataset, relinquishing any underlying resources.
+	 * <p>
+	 * For example, this would close any open file and network streams and free
+	 * database locks held by the dataset implementation.
+	 * <p>
+	 * The behaviour of the other dataset methods are undefined after closing
+	 * the dataset.
+	 * <p>
+	 * Implementations might not need {@link #close()}, hence the default
+	 * implementation does nothing.
+	 */
+	@Override
+	default void close() throws Exception {
+	}
+
+	/**
+	 * Get the default graph of this dataset.
+	 * <p>
+	 * The {@link Triple}s of the default graph are equivalent to the
+	 * {@link Quad}s in this Dataset which has the {@link Quad#getGraphName()}
+	 * set to {@link Optional#empty()}.
+	 * <p>
+	 * It is unspecified if modifications to the returned Graph are reflected in
+	 * this Dataset.
+	 * <p>
+	 * The returned graph MAY be empty.
+	 * 
+	 * @see #getGraph(BlankNodeOrIRI)
+	 * @return The default graph of this Dataset
+	 */
+	Graph getGraph();
+	
+	/**
+	 * Get a named graph in this dataset.
+	 * <p>
+	 * The {@link Triple}s of the named graph are equivalent to the
+	 * the Quads of this Dataset which has the
+	 * {@link Quad#getGraphName()} equal to the provided <code>graphName</code>, or
+	 * equal to {@link Optional#empty()} if the provided <code>graphName</code> is
+	 * <code>null</code>.
+	 * <p>
+	 * It is unspecified if modifications to the returned Graph are reflected in
+	 * this Dataset.
+	 * <p>
+	 * It is unspecified if requesting an unknown or empty graph will return
+	 * {@link Optional#empty()} or create a new empty {@link Graph}.
+	 * 
+	 * @see #getGraph()
+	 * @see #getGraphNames()
+	 * @param graphName
+	 *            The name of the graph, or <code>null</code> for the default
+	 *            graph.
+	 * @return The named Graph, or {@link Optional#empty()} if the dataset do
+	 *         not contain the named graph.
+	 */	
+	Optional<Graph> getGraph(BlankNodeOrIRI graphName);	
+	
+	/**
+	 * Get the graph names in this Dataset.
+	 * <p>
+	 * The set of returned graph names is equivalent to the set of unique
+	 * {@link Quad#getGraphName()} of all the {@link #stream()} of this
+	 * dataset (excluding the default graph).
+	 * <p>
+	 * The returned {@link Stream} SHOULD NOT contain duplicate graph names.
+	 * <p>
+	 * The graph names can be used with {@link #getGraph(BlankNodeOrIRI)} to
+	 * retrieve the corresponding {@link Graph}, however callers should be aware
+	 * of any concurrent modifications to the Dataset may cause such calls to
+	 * return {@link Optional#empty()}.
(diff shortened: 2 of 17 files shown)


stain:
Q: Should `Dataset` have a `public Graph getDefaultGraph();` method? If so - would that then need to be dynamically updating Dataset?  Or is `getTriples()` sufficient?

afs:
What is the use case for TripleOrQuad?

The proposal (the reply to the discussion referenced) is `Quad.asTriple` to go between the Quad world and the Triple world.

stain:
`TripleOrQuad` is only useful as a duck-type interface, if someone needs to be able to do `getSubject()`,  `getPredicate()` or `getObject()` on both a Triple or Quad - e.g. in a filter used with a Stream without needing to do a double `instanceof`. 

It is therefore also used in the other duck-type interface `GraphOrDataset` - they are basically most useful if they both exist - and mostly for any intermediate util methods that don't care too much if something is a Quad or Triple - rather than for end users.

You can imagine a JSON-LD serializer being able to deal with `TripleOrQuad` directly and just do the odd `instanceof Quad` check to access `getGraphName()` - otherwise it would need to have duplicates of everything.

stain:
If we have `Quad.asTriple` - then should there also be a `Triple.asQuad()`?

If so - would this then be used to get the 'original quad'?:

```
Quad q1;
assert(q1.getGraphName().isPresent());
assert(q1.asTriple().asQuad().equals(q1));
```

Otherwise - if we do add `Triple.asQuad()` as well, the should the returned Quad always be in the default graph?

stain:
BTW - I'm OK with removing both `GraphOrDataset`  and `TripleOrQuad` if we don't think it's useful to have common interface for the common methods of `Graph`/`Dataset`  and `Triple`/`Quad`.

(better names might be `GraphLike` and `TripleLike` ?)

afs:
"default graph" is a context-sensitive attribute.  Freestanding `Quad` (like a `Triple`) can be processed in ways that don't have the idea of dataset. I think this is more useful building block to have a general pruprose quad rather than than putting the idea that quads with no graph name are the default graph. A dataset is a set of triples and a set of quads.

stain:
OK, so I think what you are saying is also in line with the different kinds of semantics explained in https://www.w3.org/TR/2014/NOTE-rdf11-datasets-20140225/ - so we don't want to make too many assumptions about the default graph - and so can't automatically have a Dataset.getGraph() -- and also this implies that `GraphOrDataset<T>` interface might not be very useful as its Triple methods should be removed and it remains just as duck-typing interface.

stain:
BTW - what is your view @afs on how we should represent the default graph in a `Quad`?  Is `Optional` then meaningful - or is it better to have some kind of constant?

afs:
A dataset is a default graph and zero or more `(name, graph)` (= named graph).  Quads are nto actually mentioned in the dataset definition but following form the defn, a quad is a `(name, triple)` and many systems treat that as a 4-tuple.

`Dataset.getGraph()` returns the default graph (which must exist BTW).

`Dataset.getGraph(name)` follows the dataset definition closely.

`getQuads` could cover just named graphs or named graphs and default graph with some marker (to me, unset is better than a URI - it's more a design choice than spec following).

I'm not seeing much use for `GraphOrDataset<T>` -- is there a concrete use case?

stain:
I think if (as in the branch now) we leave out the getTriples() and other
default graph operations from Dataset, then the use case for GraphOrDataset
is dropping by the hour, as those would be accessible from (presumably
mutable) Dataset.getGraph() (or getDefaultGraph())

GraphOrDataset could still be useful in parsers (e.g. intoGraphOrDataset()
or return type) - or internally in utility code that just passes both types
around (e.g. a utility method for merging two Graphs would work just as
well iterating over a GraphOrDataset<T extends TripleOrQuad> and call
add(T).

Perhaps we should try to write a couple of such util methods just to see if
GraphOrDataset and TripleOrQuad is really needed. Generics in Java can
often just cause a bigger headache than a classic cast in the right place,
and it wouild need generics to avoid graph.add(quad)

stain:
This suggests that Quad should be containing a Triple rather the
subj/pred/obj - should asTriple () be renamed to getTriple()? The default
methods could be swapped around so default getSubject() { return
getTriple().getSubject() }

The text of Quad.hashCode() and Quad.equals() could be similarly modified
to use the getTriple(), although the actual semantics would be the same.

afs:
The interface for Quad is getGraph, getSubject, getProperty, getObject and asTriple.  Defaulting getSubject is presumimg an implementaion. As far as I know, no toolkits works that way - they tend to have Quads from parsers.

I find `asTriple` mildly better than `getTriple` - (1) too many "get"s (2) to conveys projection, a form of conversion.

On parsers: Parsers can be decoupled. They send triples/quads/prefixes/baseURIs (i.e. syntax concepts) to a channel (`StreamRDF` in Jena). Different implementations behind the channel are possble; print (i.e. parse and output NT, NQ pure streaming), send to a graph, send to a datasets (triples to default graph).

What CommonsRDF needs is that channel abstraction so external parses can be plugged in.

For parsing quads to graph there is a choice point: both make sense. Either triples to graph (i.e. extract the default graph) or all quads to triples (caveat: quads might be separate triples for a reason).
 from graph, dataset whatever.  They can be used without such as storage.

stain:
I've removed the `GraphOrDataset` superinterface, and added: 
* `Graph getGraph()`
* `Optional<Graph> getGraph(graphName)`
* `Stream<BlankNodeOrIRI> getGraphNames()`

See http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Dataset.html#getGraph-- and below

In simple's DatasetImpl these are [pretty simple](https://github.com/apache/incubator-commonsrdf/commit/839a6425fdb4985ef1594af87c6e12dcef8ec519#diff-1) for now - just extracting them from its internal `Set<Quad>` - meaning that the returned `Graph` would not mutate the Dataset.

Do we want to add stronger links between the returned Graphs and the Dataset? I can see this expanding with `addGraph` and `removeGraph`..

(An alternative DatasetImpl would be to either have a GraphInDataset proxy class, allowing mutation, or to change the backing to be a Map<BlankNodeOrIRI, Graph> -- that would complicate many of the methods to have to merge the corresponding streams from the underlying Graphs)

stain:
OK, @afs - so we'll leave `Quad` as proposed now then.

As for parsing - it might make sense to expand/split the `parser-writer-interface` branch to have a separate "low-level" parser interface that just yield triples/quads.  

Would a `Stream<T extends TripleOrQuad>` be sufficient? Implementation wise (e.g. adapting Jena's `StreamRDF`) I am not sure how easy it would be to make a Stream from a callback - perhaps using a `Queue` as intermediary?  Would love having the `yield` keyword in Java.. :)

Unlike `StreamRDF` you would lose namespace prefix bindings perhaps - but that's not really part of RDF Core concepts.

Let's continue the parser discussion on a new pull request #21

stain:
Test comment to see if it is forwarded to dev@commonsrdf

stain:
Moving forward with quads! I have taken a stance.

Rather than `TripleOrQuad` I propose a similar interface, called [TripleLike](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/TripleLike.html). This can also support generalized RDF statements with any `RDFTerm` subtype for `getSubject()`, `getPredicate()` and `getObject()`. This can be useful both for parsers and for SPARQL compatibility.

Then naturally we also have its extension [QuadLike](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/QuadLike.html) which adds a generalized `getGraphName`.

Instead of `DatasetOrGraph` I propose a generalized [GraphLike](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/GraphLike.html) that contains `TripleLike` statements. `GraphLike` do not have the split-up methods for subject/predicate/object - thus `Quad` does not need to support the ambiguous `contains(s,p,o)`.

As `Graph` and `Dataset` previously differed in returning the stream by `getTriples()` and `getQuads()`  and this crucial method therefore could not be in the common interface, I've gone brave here and renamed both to [.stream()](http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Graph.html#stream--).

What do you think?

afs:
See my previous question - what has fundamentally changed from before around equality?

When are two quad objects `.equals`?

stain:
Tracked in Jira as https://issues.apache.org/jira/browse/COMMONSRDF-37

stain:
@afs I've updated the quad javadocs to clarify my view:

http://stain.github.io/incubator-commonsrdf/quad/org/apache/commons/rdf/api/Quad.html#equals-java.lang.Object-

says:

> Two Quads are equal if and only if their getGraphName(), getSubject(), getPredicate() and getObject() are equal.
> Implementations MUST also override hashCode() so that two equal Quads produce the same hash code.
> Note that a Quad MUST NOT be equal to a Triple, even if this Quad's getGraphName() is Optional.empty(). To test triple-like equivalence, callers can use:

```java
 Quad q1;
 Triple t2;
 q1.asTriple().equals(t2));
 ```

The last statement means that `Triple` don't need to support the inverse `.equals(Quad)` and if someone really wanted to do `Quad extends Triple` then they would need to have a boolean isQuad; or similar to know which semantics to use in its `.equals()` (just as if they wanted to implement both Literal and IRI).

stain:
Doing the [rdf4j implementation](https://issues.apache.org/jira/browse/COMMONSRDF-35?) I noticed that there a [Statement](http://rdf4j.org/javadoc/latest/org/eclipse/rdf4j/model/Statement.html) has 0 or more contexts (e.g. graph names if assuming a Dataset) - however these Statements are used both in "union Graph like" and "Dataset like" settings and thus would have unclear equality semantics, in particular for `.remove()` methods.

So I think with Commons RDF you would have to force either view on the Statement rather than be a fuzzy hybrid. In fact from a single Statement you would get a single Triple, but potentially multiple Quads with different `getGraphName()`, which obviously should not be `.equals()` each-other even if they are backed by the very same Statement.

Thus for rdf4j I would assume that a a `graph.remove(triple)` would remove the `Statement` - removing it from the union graph would mean removing from all contexts.

while I would expect  `dataset.remove(quad)`  would remove that quad from only that context - that is the other quads from the other contexts would remain in the dataset.

I don't think rdf4j supports such a "Statement-modifying" call and would have to re-insert the `Statement` with the shrunk context list. Devil's in the detail.. but I can see why formalizing this kind of semantics could be valuable over "Look at the source code" for knowing what happens.

stain:
btw.. neither `TripleLike` or `QuadLike` have defined equality semantics (beyond Object.equals()) - so I've clarified that in [`GraphLike`](file:///home/stain/src/incubator-commonsrdf/target/site/apidocs/org/apache/commons/rdf/api/GraphLike.html) as that would obviously affect its `.contains()` and `.remove()` methods:

> As `TripleLike` do not have a specific `Object.equals(Object)` semantics, the behaviour of methods like `contains(TripleLike)` and `remove(TripleLike)` is undefined for arguments that are not object identical to previously added or returned `TripleLike` statements.

afs:
Looks OK - the hashCode algorithm must be defined. Both `Quad` and `Triple` could provide a hashCode of `Objects.hash(...)` as a default method.

stain:
If there are no objections, I'll merge this `Quad` and `Dataset` branch next week, then perhaps we can try to make a `0.3.0` release.

PR #20, opened 2016-04-08
Added SimpleRDFTerm marker interface

stain:
Now internallyMap() will fail early on mapping
any RDFTerm objects which are not IRI, BlankNode or Literal
(before they would have been added 'as-is')

This also fixes a bug in that Graph would have remapped
any of the Types instances to an IRIImpl.

I added SimpleRDFTerm interface as a package-protected
interface within SimpleRDFTermFactory - thus it should not
be easy to create any additional SimpleRDFTerm subclasses.


stain:
As this is in simple it should not be controversial, given 48h I'll now merge it in.

PR #21, opened 2016-04-13
Add RDFParserBuilder interface

stain:
This pull request proposes a `RDFParserBuilder` interface - a "Builder for parsing an RDF source into a Graph".

The javadocs explain the approach:

* http://stain.github.io/incubator-commonsrdf/rdfparserbuilder/org/apache/commons/rdf/api/RDFParserBuilder.html
* http://stain.github.io/incubator-commonsrdf/rdfparserbuilder/org/apache/commons/rdf/simple/AbstractRDFParserBuilder.html

`AbstractRDFParserBuilder` is here part of `simple` (but has no implementation in there as Simple has no parsers) - and basically takes care of handling the 'builder' aspect in an immutable way - but also does some more generic checking and thread handling which implementations may override.

It could be that these two aspects of AbstractRDFParserBuilder could be split, and a neutral bean-like builder abstract class could be added to `api`, and the more checker-helper-things to another abstract subclass in `simple`.

I think once we agree on the style it's easier to do the equivalent Writer interface - but feel free to contribute to the `parser-writer-interface` branch!

Example implementations:

- [jsonld branch](https://github.com/apache/incubator-commonsrdf/blob/jsonld-java/jsonld-java/src/main/java/org/apache/commons/rdf/jsonldjava/JsonLdParserBuilder.java)
- [stain/jena branch](https://github.com/stain/incubator-commonsrdf/blob/jena/jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFParserBuilder.java)  (builds on Andy's https://github.com/afs/commonsrdf-jena)

Example usage:

```java
IRI iri = factory.createIRI("http://example.com/example.jsonld");
Graph g = new JsonLdParserBuilder()
				.contentType(RDFSyntax.JSONLD).source(iri).parse()
				.get(10, TimeUnit.SECONDS);
```

Some things to discuss:

- Should this merge the `quad` branch from #19 and add `intoDataset(Dataset)`?  Would the return type from `parse()` then be a `Future<GraphOrDataset>`? (Or would a `ParseResult` wrapper be better?)
- How can `RDFParserBuilder` implementations be discovered? Another `ServiceLoader`, or part of `RDFTermFactory` (which then need to change name to just `RDFFactory` ?)
- Should there be a kind of `canHandle(RDFSyntax)` method, or would you ask for the parser with `rdfFactory.getParser(RDFSyntax.JSONLD)` instead?  (Note: sometimes the content type is not known in advance, e.g. URIs on the internet)
- Should there be a concrete implementation that calls down to more lower-level parser interfaces? E.g. equivalent to Jena's `RDFDataMgr`?

Note - my branch with jsonld-java is a bit of an outlier as it's not a complete RDF framework and mainly does JSON-LD - however N-Quad and Turtle support is already part of jsonld-java - however I'm not sure if that is just for parsing or writing.  We would expect most of the bigger frameworks to do most of the standard syntaxes we enumerated in [RDFSyntax](https://commonsrdf.incubator.apache.org/apidocs/org/apache/commons/rdf/api/RDFSyntax.html)

### DIFF
--- a/api/src/main/java/org/apache/commons/rdf/api/RDFParserBuilder.java
+++ b/api/src/main/java/org/apache/commons/rdf/api/RDFParserBuilder.java
@@ -34,22 +34,34 @@ import java.util.concurrent.Future;
  * {@link #source(InputStream)}) before calling {@link #parse()} on the returned
  * RDFParserBuilder - however methods can be called in any order.
  * <p>
+ * The call to {@link #parse()} returns a {@link Future}, allowing asynchronous parse
+ * operations. This can be combined with {@link #intoGraph(Graph)}
+ * allowing access to the graph before parsing has completed,
+ * however callers are still recommended to to check 
+ * {@link Future#get()} for any exceptions thrown during parsing.
+ * <p>
  * Setting a method that has already been set will override any existing value
  * in the returned builder - irregardless of the parameter type (e.g.
  * {@link #source(IRI)} will override a previous {@link #source(Path)}. Settings
  * can be unset by passing <code>null</code> - this may require casting, e.g.
- * <code>contentType( (RDFSyntax) null )</code>.
+ * <code>contentType( (RDFSyntax) null )</code> to undo a previous call to
+ * {@link #contentType(RDFSyntax)}.
  * <p>
  * It is undefined if a RDFParserBuilder is mutable or thread-safe, so callers
  * should always use the returned modified RDFParserBuilder from the builder
  * methods. The builder may return itself, or a cloned builder with the modified
- * settings applied.
+ * settings applied. Implementations are however encouraged to be 
+ * immutable and thread-safe and document this, 
+ * as an example starting point, see 
+ * {@link org.apache.commons.rdf.simple.AbstractRDFParserBuilder}.
  * <p>
  * Example usage:
  * </p>
  * 
  * <pre>
- *   Graph g1 = ExampleRDFParserBuilder.source("http://example.com/graph.rdf").parse();
+ *   // Retrieve populated graph from the Future
+ *   Graph g1 = ExampleRDFParserBuilder.source("http://example.com/graph.rdf").parse().get(30, TimeUnit.Seconds);
+ *   // Or parsing into an existing Graph:
  *   ExampleRDFParserBuilder.source(Paths.get("/tmp/graph.ttl").contentType(RDFSyntax.TURTLE).intoGraph(g1).parse();
  * </pre>
  * 
@@ -138,6 +150,15 @@ public interface RDFParserBuilder {
 	 * {@link #parse()} return a new {@link Graph}, which is created using
 	 * {@link RDFTermFactory#createGraph()} if
 	 * {@link #rdfTermFactory(RDFTermFactory)} has been set.
+	 * <p>
+	 * It is undefined if any triples are added to the specified
+	 * {@link Graph} if the {@link Future#get()} returned from 
+	 * {@link #parse()} throws any exceptions. (However 
+	 * implementations are free to prevent this using transaction 
+	 * mechanisms or similar).  However, if {@link Future#get()}
+	 * does not indicatean exception, the 
+	 * parser implementation SHOULD have inserted all parsed triples 
+	 * to the specified graph.
 	 * 
 	 * @param graph
 	 *            The {@link Graph} to add triples into.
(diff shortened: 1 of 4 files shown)


stain:
Already by implementing this for [jsonld-java](https://github.com/apache/incubator-commonsrdf/blob/jsonld-java/jsonld-java/src/main/java/org/apache/commons/rdf/jsonldjava/JsonLdParserBuilder.java), I kind of struggled with the assumptions I made on how easy it would be to stream triples across - jsonld-java insists on making its own `RDFDataset` in memory - fine - as we simply wrap it in a [JsonLdGraph](https://github.com/apache/incubator-commonsrdf/blob/jsonld-java/jsonld-java/src/main/java/org/apache/commons/rdf/jsonldjava/JsonLdGraph.java)

..but now this fights against the `AbstractRDFParserBuilder.parseSynchronously()` interface, as it still has to clone it into the `intoGraph()` even when it's a brand new `JsonLdGraph`.  (It can however [use .putAll()` if the target is empty](https://github.com/apache/incubator-commonsrdf/blob/jsonld-java/jsonld-java/src/main/java/org/apache/commons/rdf/jsonldjava/JsonLdParserBuilder.java#L98)) 

Part of this might be issues with the API of jsonld-java - e.g. there's a `TripleCallback` - but it doesn't give you triples as they are found, but a complete `RDFDataset`

stain:
@afs [mentioned](https://github.com/apache/incubator-commonsrdf/pull/19#issuecomment-208531395) that it could be good to have a kind of streaming/channel interface for parsers and writers that is not bound to any Graph or Dataset - something similar to Jena's [StreamRDF](https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/riot/system/StreamRDF.html) perhaps - or just a `Stream<TripleOrQuad>`?

Would that be instead of the approach proposed in this PR, or in addition?

afs:
There would not be a `Future<Graph> parse()` because the destination may not be a graph.

It would be

    void setOutput(StreamRDF)
    void parse()

The common case is parsers that push output.

A pull-parser design could be done but the interface would not be a single `Stream<TripleOrQuad>` because a  parser can produce triples and quads. A single stream design moves the burden of testing the type onto the consumer. 

A pull-parser can be built on top of `setOutput`/`parse` by using channels.

stain:
How about (calling it `target` rather than `output`):

```
RDFParserBuilder tripleTarget(Consumer<Triple> tripleConsumer);

default RDFParserBuilder tripleTarget(Graph graph) {
		return target(graph::add);
}

RDFParserBuilder quadTarget(Consumer<Quad> quadConsumer);
default RDFParserBuilder quadTarget(Dataset dataset) {
		return target(dataset::add);
}
```

I am not so sure about replicating [Jena StreamRDF](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/riot/system/StreamRDF.java)'s mix of `triple(Triple)` and `quad(Quad)` - as it is confusing as if triples in the default graph will appear in quad, triple or both.

(I know I'm introducing the same confusion above..)

Perhaps if we accept the `Quad` interface, then the parsers can always deal with Quads - and thus callers don't need to hand-held too much based on content types.

stain:
Hm, I'm trying now with 

```java
RDFParserBuilder target(Consumer<Quad> consumer);

default RDFParserBuilder target(Graph graph) {		
		return target(q -> { 
			if (! q.getGraphName().isPresent()) { 
				graph.add(q.asTriple());
			}
		});
	}

default RDFParserBuilder target(Dataset dataset) {
		return target(dataset::add);
	}

public interface ParseResult {}

Future<? extends ParseResult> parse()
```

(Where `ParseResult` can be `null`, or a subinterface defined by the implementation with debug info)

but then this means that the parser can no longer have an efficient default of using its own Graph/Dataset implementation. I don't see why someone using a parser needs to decide on the graph implementation - it makes the parserbuilder interface just a bit cumbersome to use as you then are forced to pick a `RDFTermFactory` (and thus should also set it with `builder.rdfTermFactory()`)

Would it make sense to extend ParseResult to have Optional<Graph> and Optional<Dataset> ?  These would be populated if `target(null)` (e.g. the default) - or also when `target(graph)` or `target(dataset)` has been set.

afs:
We have `FactoryRDF` for deciding graph implementations.

Humbedooh:
just testing that comments make it to the list, ignore :)

PR #23, opened 2016-09-28
COMMONSRDF-33 Add Jena integration

stain:
This adds the `jena` module, Commons RDF implementations backed by [Apache Jena](http://jena.apache.org/).

See also [COMMONSRDF-33](https://issues.apache.org/jira/browse/COMMONSRDF-33)


stain:
Here is what the [combined Javadoc](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/) looks like, including [org.apache.commons.rdf.jena](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/jena/package-summary.html)

Worthy of review is perhaps all the method variations in [JenaRDFTermFactory](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/jena/JenaRDFTermFactory.html).

jena/src/main/java/org/apache/commons/rdf/jena/JenaQuadLike.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.apache.commons.rdf.jena;
+
+import org.apache.commons.rdf.api.QuadLike;
+import org.apache.commons.rdf.api.RDFTerm;
+import org.apache.jena.sparql.core.Quad;
+
+/** Access the Jena quad backing this object */
+public interface JenaQuadLike<S extends RDFTerm, P extends RDFTerm, O extends RDFTerm, G extends RDFTerm>
```
ajs6f: Is this the generalized-RDF quad?

jena/src/main/java/org/apache/commons/rdf/jena/JenaGraph.java
```diff
@@ -0,0 +1,45 @@
...
+	 * Return the underlying Jena {@link Graph}.
+	 * <p>
+	 * Changes to the Jena graph are reflected in the Commons RDF graph and vice versa.
+	 * 
+	 * @return A Jena {@link Graph}
+	 */
+	public Graph asJenaGraph();
+
+	/**
+	 * Return the underlying Jena {@link Model}.
```
ajs6f: There isn't always an underlying `Model` in the impl below; sometimes one is created to wrap the underlying Jena `Graph`, which is what is available. I think this comment could be clearer on that point, because Jena keeps `Node`/`Triple`/`Graph`/`DatasetGraph` and `RDFNode`/`Statement`/`Model`/`Dataset` pretty distinct.

jena/src/main/java/org/apache/commons/rdf/jena/JenaQuad.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.apache.commons.rdf.jena;
+
+import org.apache.commons.rdf.api.BlankNodeOrIRI;
+import org.apache.commons.rdf.api.IRI;
+import org.apache.commons.rdf.api.RDFTerm;
+
+/** Access the Jena quad backing this object */
+public interface JenaQuad extends org.apache.commons.rdf.api.Quad,
```
ajs6f: Is this the defined point to expose Jena's generalized-RDF quad?

jena/src/main/java/org/apache/commons/rdf/jena/JenaQuadLike.java
ajs6f: Is this the defined point to expose Jena's generalized-RDF quad?

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTerm.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.commons.rdf.jena;
+
+import org.apache.commons.rdf.api.RDFTerm;
+import org.apache.jena.graph.Node;
+
+/** Access the Jena node backing this object */
+public interface JenaRDFTerm extends RDFTerm {
```
ajs6f: Did you think about making this generic i.e.
```
public interface JenaRDFTerm<NodeType extends Node> extends RDFTerm {
        public NodeType asJenaNode();
```
and
```
class AbstractJenaRDFTerm<NodeType extends Node> implements JenaRDFTerm<NodeType>, RDFTerm {
+	private NodeType node;
```
and then for example:
```
public interface JenaIRI extends JenaRDFTerm<Node_URI>, IRI {
```
etc. That's more verbose, but maybe a bit safer?

ajs6f:
It seems like the generalized quad is supported, but not the triple. Am I missing something?

jena/src/main/java/org/apache/commons/rdf/jena/JenaGraph.java
stain: OK, so you think we should keep the method without needing `Optional`, but make it clear that it might be made on-demand.

ajs6f: Yes, that seems a bit clearer. It's really the word "underlying" (which seems to imply preexisting) that gave me pause. But I'd like to hear from other, more experienced Jena committers. I may be emphasizing the separation between the SPI and the API too strongly.

jena/src/main/java/org/apache/commons/rdf/jena/JenaQuadLike.java
stain: Yes, but perhaps there should be a subtype `JenaGeneralizedQuadLike`?  I was reluctant to imply it was a  `Quad` subtype, as I think it should not have the `Quad.equals()` logic.

Now [fromJenaGeneralized()](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/jena/JenaRDFTermFactory.html#fromJenaGeneralized-org.apache.jena.sparql.core.Quad-) returns a ```JenaQuadLike<RDFTerm,RDFTerm,RDFTerm,RDFTerm>```

I see now that I have [createGeneralizedTriple()](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/jena/JenaRDFTermFactory.html#createGeneralizedTriple-org.apache.commons.rdf.api.RDFTerm-org.apache.commons.rdf.api.RDFTerm-org.apache.commons.rdf.api.RDFTerm-) but forgot `createGeneralizedQuad()` - obviously both should be there.

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTermFactory.java
```diff
@@ -0,0 +1,874 @@
...
+		BlankNodeOrIRI subject;
+		IRI predicate;
+		try {
+			subject = (BlankNodeOrIRI) fromJena(factory, triple.getSubject());
+			predicate = (IRI) fromJena(factory, triple.getPredicate());
+		} catch (ClassCastException ex) {
+			throw new ConversionException("Can't convert generalized triple: " + triple, ex);
+		}
+		RDFTerm object = fromJena(factory, triple.getObject());
+		return factory.createTriple(subject, predicate, object);
```
ajs6f: Just as a style-thing, it would seem nicer to me to put everything (declaration, assignment, `return`) inside the `try`, i.e.
```
    try {
	    BlankNodeOrIRI subject = (BlankNodeOrIRI) fromJena(factory, triple.getSubject());
	    IRI predicate = (IRI) fromJena(factory, triple.getPredicate());
	    RDFTerm object = fromJena(factory, triple.getObject());
	    return factory.createTriple(subject, predicate, object);
    } catch (ClassCastException ex) {
	    throw new ConversionException("Can't convert generalized triple: " + triple, ex);
    }
```

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTerm.java
stain: Yeah.. I didn't initially see the big win for that, as you would typically have to do `instanceof JenaBlankNode` etc. anyway in most cases - but I agree that is safer, and would avoid an anti-pattern like:

```
RDFTerm term; // ..
if (term instance of JenaBlankNode) {
JenaBlankNode bnode = (JenaBlankNode)term;

// anti-pattern - can't know this for sure:
Node_URI uri = (Node_URI)bnode.asJenaNode();
}
```

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTermFactory.java
stain: I dare to disagree in this instance as `ClassCastException` is unchecked - - the try-catch should only cover what we want to catch.

Some might argue against a `return` from inside a `try` block, but I don't see a problem with that side (Java won't let you void through).

We don't expect `fromJena()` or `createTriple()` to throw ClassCastException. If they do, it should be a proper `RuntimeException`, not a `ConversionException`.

stain:
Thanks for the review, @ajs6f! I'll try to fix.

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTermFactory.java
ajs6f: Okay, fair enough. I was too concerned with concision. I think it should be possible to make the variable `final`, though, to indicate that there is only one correct path through.

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTerm.java
ajs6f: Yeah, it's not a huge thing. It's a question of clear intent and yes, that second form is icky. You'd rather not see that. But without some common enumeration of the possible term types, there's no avoiding instance checking. This is where I bet you'd like to be writing Scala. :)

ajs6f:
Glad to, @stain ! This is good stuff, and again, let me know if/how I can help.

jena/src/main/java/org/apache/commons/rdf/jena/JenaRDFTerm.java
stain: Another problem with typing this I see now are some of the more special Node types in Jena that confuses me:

`Node_Blank`, `Node_NULL` etc.  


We don't use the specific Node interfaces within this integration, as `Node` itself has the required `.isBlank()` etc methods - I don't want to second-guess if `.isBlank()` == true means it has to be a `Node_Blank` instance; this sounds more like an internal interface.

stain:
OK, pushed fixes. I didn't fix the `<T extends Node>` because of the weird Node sub-types, and I didn't add a `JenaGeneralizedQuadLike`.

stain:
I also exposed `createdAnyVariable()` and `createVariable(String)` in the `JenaRDFTermFactory` as those were hidden inside `impl.JenaFactory` - not sure if those would be useful with a generalized triple/quad or only would appear in a SPARQL query?

ajs6f:
The "weird" Node sub-types question is a good one. My sense is that they are indeed internal. @afs , can you comment here? I like to type as strongly as possible, but I don't want to lead anyone astray! :)

ajs6f:
The comment on Jena's `Node_Variable`:
```
    "variable" nodes; these are outside the RDF2003 specification, but are
    used internally for "placeholder" nodes where blank nodes would be
    wrong, most specifically in Query.
```
leads me to believe that we are talking about SPARQL here. There's not really any other place in Jena I know about that would use such machinery-- the `find` methods on various structures just use `null` or `Node_ANY` as a wildcard.

ajs6f:
Also, the phrase "used internally" in that comment makes me wonder if it is right to expose those methods?

stain:
So I guess they would be mainly to represent the "triple patterns" in SPARQL, e.g. `?s` in:

```turtle
?s a <http://example.com/>;
```

although there would not be any way to use those jena variables with the Commons RDF-exposed graph/datasets - and they should not be returned in there. (unless someone had deliberately added such to a Jena `Graph` manually - which is allowed today as it excepts any `Node`)

Do I get it right that these would not be appearing in query *results*?

ajs6f:
I don't think there is any situation in which a variable can appear as part of query results (except, I suppose, if you want to think about the label of a variable appearing as the name of a column).

stain:
OK, then I'll remove `JenaVariable` and merge the rest. Thanks for the review, @ajs6f !

stain:
Hmm.. Removing `JenaVariable` means in `fromJenaGeneralized(Node)` I have to fail on `node.isVariable()` (which [is in the Jena API](https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/Node.html#isVariable--)):

e.g. current suggestion

```java
if (node.isVariable()) {
  return new JenaVariableImpl(node);
}
```

will instead become:

```java
throw new IllegalArgumentException("Unrecognized node type: " + node);
```

@afs  - how generalized should I make the generalized quad? :)

afs:
I'd keep to concrete RDFTerms, i.e. no named variables.  Interop is not served by having them.

ajs6f:
Instead of
```
throw new IllegalArgumentException("Unrecognized node type: " + node);
```
what about 
```
throw new ConversionException(new UnsupportedOperationException("Conversion does not currently support SPARQL variables : " + node));
```

jena/src/main/java/org/apache/commons/rdf/jena/ConversionException.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Exception thrown when a problem arises across the Jena-CommonRDF boundary.
+ * This should not happen in well-formed RDF data but, for example, Jena triples
+ * allow
+ * <a href="http://www.w3.org/TR/rdf11-concepts/#section-generalized-rdf" >
+ * generalized RDF</a>.
+ */
+
+public class ConversionException extends JenaException {
+	public ConversionException() {
```
afs: As this is thrown going from Jena to Commons RDF, shouldn't be a CommonsRDF exception?  i.e. the app code does "try { ... } catch (CommonsRDFException)" not "catch (JenaException)" because it's jena-independent at that point?

afs: Not sure there can be a conversion error from commons to Jena.  It is the "not RDF Terms" like Node.ANY, and named variables, that can trigger ConversionException or out-of-place literals.  I may have missed any in the other direction.  

For CommonsRDF applications, exposure to implementation-related exceptions seems wrong.

There is at least one throws declarations but not most places - as this is unrecoverable, style suggests no declaration. YMMV.

stain: OK, I will just make `ConversionException` be a `RuntimeException` instead of involving `JenaException`. However I guess there could still be `JenaException` thrown from inside many of these methods - but if that is likely to happen we should say so explicitly.

jena/src/main/java/org/apache/commons/rdf/jena/JenaAny.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.commons.rdf.jena;
+
+import org.apache.jena.graph.Node;
+
+/**
+ * A placeholder for "any" RDFTerm.
+ * <p>
+ * Equivalent to <code>[]</code> in SPARQL. This is a Jena-specific
+ * RDFTerm extension used in SPARQL queries.
```
stain: OK, I've removed both `JenaVariable` and `JenaAny` (our APIs from the `api.Graph` and `.api.Dataset` says to use `null` in patterns, e.g. `graph.contains(s,null,p)`

stain:
Thanks for the great review, @ajs6f  and @afs! Now merged in `master`.

ajs6f:
Great stuff, @stain -- thanks for moving this forward!

jena/src/main/java/org/apache/commons/rdf/jena/ConversionException.java
afs: I don't think you'll get an JenaExceptions (unless things are really quite  broken).

PR #24, opened 2016-09-28
COMMONSRDF-35 Add rdf4j integration

stain:
This adds the `rdf4j` module, adding integrations for [Eclipse RDF4J](http://rdf4j.org/) - formerly known as Sesame.


stain:
Here is what the [combined Javadoc](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/) looks like, including [org.apache.commons.rdf.rdf4j](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/rdf4j/package-summary.html)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
```diff
@@ -0,0 +1,134 @@
...
+		model.add(rdf4jTermFactory.asStatement(triple));
+	}
+
+	public Optional<Model> asModel() { 
+		return Optional.of(model);
+	}
+
+	@Override
+	public Optional<Repository> asRepository() {
+		return Optional.empty();
```
ansell: As above, I would prefer this throw an exception to indicate that it may function in future when someone implements a ModelRepository in RDF4J (or Marmotta sends their implementation upstream), but it currently doesn't actually work. Using Optional is not a good substitute for UnsupportedOperationException in my view.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
```diff
@@ -0,0 +1,134 @@
...
+	}
+	
+	@Override
+	public void clear() {
+		model.clear();
+	}
+
+	@Override
+	public boolean contains(BlankNodeOrIRI subject, org.apache.commons.rdf.api.IRI predicate, RDFTerm object) {
+		return model.contains(
```
ansell: Does this method currently work for all BlankNodes (or is it documented to be precarious in those situations)? Ie. is it tested to work in all of the various corner cases for BNode to BlankNode conversion/wrapping or are their known and declared cases where it isn't expected to work consistently.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+	@Override
+	public long size() {
+		int size = model.size();
+		if (size < Integer.MAX_VALUE) {
+			return size;
+		} else {
+			// TODO: Check if this can really happen with RDF4J models
+			// Collection.size() can't help us, we'll have to count
+			return model.parallelStream().count();
```
ansell: I doubt it is possible or practical for in-memory implementations of Model, but going over Integer.MAX_VALUE definitely occurs for Repository in some cases.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
```diff
@@ -0,0 +1,134 @@
...
+				(org.eclipse.rdf4j.model.IRI)rdf4jTermFactory.asValue(predicate), 
+				rdf4jTermFactory.asValue(object)).parallelStream()
+			.map(rdf4jTermFactory::asTriple);
+	}
+	
+	@Override
+	public Optional<Resource[]> getContextFilter() {
+		// ModelGraph always do the unionGraph
+		return Optional.empty();
+		// TODO: Should we support contextFilter like in RepositoryGraphImpl?
```
ansell: To be consistent, the mask/filter should be implemented the same for both the Model and Repository versions. Model supports the same methods for context selection as Repository so it shouldn't be difficult to implement them both.

ansell: There may be some complexity added to some of the stream based methods above if context selection is supported though, so there are tradeoffs.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
```diff
@@ -0,0 +1,192 @@
...
+		} else {	
+			BlankNodeOrIRI g = graphName.orElse(null);
+			Resource context = (Resource) rdf4jTermFactory.asValue(g);
+			contexts = new Resource[] { context };
+		}
+		return contexts;
+	}
+
+	@Override
+	public void remove(Optional<BlankNodeOrIRI> graphName, BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
```
ansell: Personally I would prefer that this method accepted arbitrary numbers of graphs at the same time, but it is an override from the API so that isn't necessary for this pull request. 

I have designed applications around the use of Resource... varargs with multiple graphs needing to be searched/updated for each case in the past that would be slightly more complex if I needed to enumerate every possible graph that I was dealing with for the current query.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
```diff
@@ -0,0 +1,192 @@
...
+		Resource subj = (Resource) rdf4jTermFactory.asValue(subject);
+		org.eclipse.rdf4j.model.IRI pred = (org.eclipse.rdf4j.model.IRI) rdf4jTermFactory.asValue(predicate);
+		Value obj = rdf4jTermFactory.asValue(object);
+		Resource[] contexts = asContexts(graphName);
+
+		try (RepositoryConnection conn = getRepositoryConnection()) {
+			conn.remove(subj, pred, obj, contexts);
+			conn.commit();
```
ansell: committing after removing a single triple is likely to cause large performance issues in the long term.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
```diff
@@ -0,0 +1,192 @@
...
+	@Override
+	public Stream<RDF4JQuad> stream(Optional<BlankNodeOrIRI> graphName, BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
+		Resource subj = (Resource) rdf4jTermFactory.asValue(subject);
+		org.eclipse.rdf4j.model.IRI pred = (org.eclipse.rdf4j.model.IRI) rdf4jTermFactory.asValue(predicate);
+		Value obj = rdf4jTermFactory.asValue(object);
+		Resource[] contexts = asContexts(graphName);
+		
+		RepositoryConnection conn = getRepositoryConnection();
+		// NOTE: connection will be closed outside by the Iterations.stream()
+		RepositoryResult<Statement> statements = conn.getStatements(subj, pred, obj, includeInferred, contexts);
```
ansell: You still need to reliably close this RepositoryResult in the case where the iteration below fails to complete. You can use try-with-resources as RepositoryResult implements AutoCloseable.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
```diff
@@ -0,0 +1,192 @@
...
+	public Optional<Graph> getGraph(BlankNodeOrIRI graphName) {
+		// NOTE: May be null to indicate default context
+		Resource context = (Resource) rdf4jTermFactory.asValue(graphName);		
+		return Optional.of(new RepositoryGraphImpl(repository, includeInferred, context));		
+	}
+
+	@Override
+	public Stream<BlankNodeOrIRI> getGraphNames() {
+		RepositoryConnection conn = getRepositoryConnection();
+		RepositoryResult<Resource> contexts = conn.getContextIDs();
```
ansell: Ditto for reliably closing this RepositoryResult object.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
```diff
@@ -0,0 +1,164 @@
...
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.repository.Repository;
+import org.eclipse.rdf4j.repository.RepositoryConnection;
+import org.eclipse.rdf4j.repository.RepositoryResult;
+
+public class RepositoryGraphImpl extends AbstractRepositoryGraphLike<Triple> implements Graph, RDF4JGraph {
+
+	private final Resource[] contextFilter;
+
+	public RepositoryGraphImpl(Repository repository, boolean includeInferred, boolean unionGraph) {
```
ansell: It may be just adding complexity for users to have this constructor. Relying on the typical RDF4J pattern in the other constructor should be sufficient.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
```diff
@@ -0,0 +1,164 @@
...
+			return conn.hasStatement(statement, includeInferred, contextFilter);
+		}
+	}
+
+	@Override
+	public void remove(Triple tripleLike) {
+		Statement statement = rdf4jTermFactory.asStatement(tripleLike);
+		try (RepositoryConnection conn = getRepositoryConnection()) {
+			conn.remove(statement, contextFilter);
+			conn.commit();
```
ansell: Ditto for comment earlier about committing after removing a single triple. Is it possible to optimise this at all?

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
```diff
@@ -0,0 +1,164 @@
...
+	}
+	
+	@Override
+	public Stream<RDF4JTriple> stream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
+		Resource subj = (Resource) rdf4jTermFactory.asValue(subject);
+		org.eclipse.rdf4j.model.IRI pred = (org.eclipse.rdf4j.model.IRI) rdf4jTermFactory.asValue(predicate);
+		Value obj = rdf4jTermFactory.asValue(object);
+		RepositoryConnection conn = getRepositoryConnection();
+		// FIXME: Is it OK that we don't close the connection?
+		RepositoryResult<Statement> statements = conn.getStatements(subj, pred, obj, includeInferred, contextFilter);
```
ansell: RepositoryResult must be closed consistently using try-with-resources rather than relying on close-on-exhaust that Iterations may trigger below.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
```diff
@@ -0,0 +1,164 @@
...
+	}
+	
+	@Override
+	protected RDF4JTriple asTripleLike(Statement statement) {
+		return rdf4jTermFactory.asTriple(statement);
+	}
+
+	public Optional<Resource[]> getContextFilter() {
+		// Make sure we clone
+		return Optional.ofNullable(contextFilter).map(f -> f.clone());
```
ansell: contextFilter should never be null so this could instead be Optional.of

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JDataset.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.commons.rdf.api.Dataset;
+import org.apache.commons.rdf.api.Quad;
+
+
+/**
+ * Marker interface for RDF4J implementations of Dataset.
+ * 
+ */
+public interface RDF4JDataset extends Dataset, RDF4JGraphLike<Quad> {
```
stain: [GraphLike](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/api/GraphLike.html) is the common interface between `Graph` and `Dataset` - operating over `TripleLike` entries.

This is mainly useful for general code that can deal with both graph and dataset (e.g. a parser/writer that iterates over `TripleLike`) - a `Graph` and `Dataset` would have different equality semantics on operations like `.contains()` and `remove()`.

We might not need a corresponding [RDF4JGraphLike](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.html) - it  exposes `.asModel()` and `asRepository()`.

Internally in `rdf4j/` there's an abstract class [AbstractRepositoryGraphLike](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.html) which shares the code of [RepositoryDatasetImpl](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.html) (implementing `Dataset`) and [RepositoryGraphImpl](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.html) (implementing `Graph`).

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraph.java
```diff
@@ -0,0 +1,66 @@
...
+	 * contexts, while retrieval (e.g. {@link #contains(Triple)}) will succeed
+	 * if the triple is in at least one of the specified contexts.
+	 * <p>
+	 * The context filter array may contain <code>null</code>, indicating the
+	 * default context (the <em>default graph</em> in RDF datasets).
+	 * <p>
+	 * If the context filter is {@link Optional#empty()}, then this is a
+	 * <em>union graph</em> which triples reflecting statements in any contexts.
+	 * Triples added to the graph will be added in the default context
+	 * <code>null</code>.
```
stain: Yes, this was very confusing from RDF4J's API as I am not sure where the nulls are allowed and how adding to the default graph should be best done - I understood it that it needs to be an a `new Resource[]{null}` array, not just `null`. 

(When calling the varargs you would need to cast as `(Resource)null` or make the array explicitly to get it right)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraph.java
```diff
@@ -0,0 +1,66 @@
...
+	 * <p>
+	 * The filter array is a copy, and thus any modifications are not reflected
+	 * in the RDF4JGraph.
+	 * 
+	 * @return The context filter as an array of {@link Resource}s, or
+	 *         {@link Optional#empty()} indicating the union graph (any
+	 *         context).
+	 */
+	public Optional<Resource[]> getContextFilter();
```
stain: Problem with `getContexts()` is that it sounds like a query (what contexts does this graph incidentally cover) rather than a selection (what contexts does it select/add to).

On that note `contextFilter` does not make sense for insertion either.. :).  Let's try `getContextMask()` then.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.repository.Repository;
+
+/**
+ * Marker interface for RDF4J implementations of GraphLike.
+ * 
+ * @see RDF4JGraph
+ * 
+ */
+public interface RDF4JGraphLike<T extends TripleLike<BlankNodeOrIRI, IRI, RDFTerm>>
```
stain: Yeah, it's tricky because Java does not have the equivalent of `owl:equivalentClass` and we wanted `TripleLike` to also be useful for generalized RDF (as in Jena - [createGeneralizedTriple](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/jena/JenaRDFTermFactory.html#createGeneralizedTriple-org.apache.commons.rdf.api.RDFTerm-org.apache.commons.rdf.api.RDFTerm-org.apache.commons.rdf.api.RDFTerm-)).

so if we make:

```
public interface RDFTripleLike extends TripleLike<BlankNodeOrIRI, IRI, RDFTerm> {}
```

then that would simplify uses like `RDF4JGraphLike` (both Graph and Dataset requires the those RDF types for s/p/o), but it would mean you could not do "upcasting" from `TripleLike`:

```
TripleLike<BlankNodeOrIRI, IRI, RDFTerm> t; // get from outside
rdfThingie.write((RDFTripleLike)t);
```

So `TripleLike` would then be a bit more pointless than it is now for most users :)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JIRI.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface RDF4JIRI 
+	extends RDF4JTerm<org.eclipse.rdf4j.model.IRI>, org.apache.commons.rdf.api.IRI {
+	
+	/**
+	 * Return the corresponding RDF4J {@link org.eclipse.rdf4j.model.IRI}.
+	 * 
+	 * @return The corresponding RDF4J IRI
+	 */
+	public org.eclipse.rdf4j.model.IRI asValue();
```
stain: `RDF4JIRI` is not an instance of `org.eclipse.rdf4j.model.IRI` **and** `org.apache.commons.rdf.api.IRI` - however it's a `RDF4JTerm` specialization with the rdf4j `IRI` as its [encapsulated field](https://github.com/apache/incubator-commonsrdf/blob/06711f9c2832609680b9775de39456b5b2b71ccb/rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRDFTerm.java).

It's an interesting idea to make `RDF4JIRI` implement both interfaces - but then we would be violating the `false` definition for Commons RDF [IRI.equals()](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-35/org/apache/commons/rdf/api/IRI.html#equals-java.lang.Object-) and [RDF4J IRI.equals()](http://rdf4j.org/javadoc/2.0/org/eclipse/rdf4j/model/IRI.html#equals-java.lang.Object-) -- perhaps both definitions could be loosened to only define the `true` case and be open-ended - but it means you could have:

```java
org.eclipse.rdf4j.model.IRI a; // from regular RDF4j
org.apache.commons.rdf.rdf4j.RDF4JIRI b; // from RDF4JTermFactory
org.apache.commons.rdf.api.IRI c; // from JenaRDFTermFactory
```

and equalities:
```java
assert a.equals(b);
assert b.equals(a);
assert b.equals(c);
assert c.equals(b);
```

but confusingly NOT

```java
org.eclipse.rdf4j.model.IRI a; // from regular RDF4j
org.apache.commons.rdf.rdf4j.RDF4JIRI b; // from RDF4JTermFactory
org.apache.commons.rdf.api.IRI c; // from JenaRDFTermFactory

assert a.equals(c);
assert c.equals(a);
```

which could lead to fun behaviour in say a HashMap..

To support it properly I think we would need to make `org.eclipse.rdf4j.model.IRI` extend `org.apache.commons.rdf.api.IRI` directly.

stain: http://www.javapractices.com/topic/TopicAction.do?Id=17 says

> 
If you extend a concrete class, and add a new field which contributes to equals, then it's not possible to write a perfectly correct equals method for the new class. Instead, you should use composition instead of inheritance. (See Effective Java by Joshua Bloch for more information.)

So in Commons RDF and RDF4J it's possible to do such inheritance because both have a defined `.equals()` - however that is also what prevents them from being directly compatible in `RDF4JIRI`. Thus `RDF4JIRIImpl` uses composition.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JLiteral.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface RDF4JLiteral 
+	extends RDF4JTerm<org.eclipse.rdf4j.model.Literal>, org.apache.commons.rdf.api.Literal {
+	
+	/**
+	 * Return the corresponding RDF4J {@link org.eclipse.rdf4j.model.Literal}.
+	 * 
+	 * @return The corresponding RDF4J Literal
+	 */
+	Literal asValue();
```
stain: Agree it is a null-override; except for the more specific Javadoc.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JQuad.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.apache.commons.rdf.rdf4j;
+
+import org.apache.commons.rdf.api.Quad;
+
+/**
+ * Marker interface for RDF4J implementations of Quad.
+ * 
+ */
+public interface RDF4JQuad extends Quad, RDF4JTripleLike {
```
stain: Both [Quad](http://stain.github.io/incubator-commonsrdf/integration/org/apache/commons/rdf/api/Quad.html) and [Triple](http://stain.github.io/incubator-commonsrdf/integration/org/apache/commons/rdf/api/Triple.html) are [TripleLike](http://stain.github.io/incubator-commonsrdf/integration/org/apache/commons/rdf/api/TripleLike.html) in CommonsRDF, they both have methods `getSubject()`, `getPredicate()` and `getObject()`.  However `Quad` and `Triple` have different equality semantics (quad also considers the graph label) - which is particularly important for a `Statement` because a Triple can be in multiple contexts.

`Quad` can't be a subtype of `Triple` as you don't get bidirectional `.equals()` (or we would have to require `Triple.equals()` to recognize `Quad`) - but they can share the "duck typing" interface `TripleLike` for users who don't care about equality, but just want to access say the `getSubject()` method. See #19 for discussion.
 
To determine if a _triple like_ is `Triple` or `Quad`, use `instanceof`. (It may be a generalized triple, which is neither).

rdf4j/pom.xml
```diff
@@ -0,0 +1,171 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-rio-trig</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-rio-jsonld</artifactId>
+		</dependency>
+		<!-- Languages notably missing, and why:
```
stain: I think there's argument either way.. we could have a pure dependency on `rdf4j-model` and `rdf4j-repository-api`- (which would minimize the chances of versioning conflicts and us needing to pull in the rdf4j bom) but then you can't parse any of the useful syntaxes. As a Commons project I think most Maven users would expect it to work "out of the box".

Perhaps the `RDF4JParser` could be in a separate Maven module (it's experimental for 0.3.0 anyway) - or we could just include instructions on how to add the different parser implementations. (I've not written docs on how to use the rdf4j or jena implementation modules yet)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+	 * unless a separate {@link RDF4JTermFactory}
+	 * instance is used per RDF4J repository/model.  
+	 * 
+	 * @see #asQuad(Statement, UUID)
+	 * @param statement
+	 *            The statement to convert
+	 * @return A {@link RDF4JQuad} that is equivalent to the statement
+	 */
+	public RDF4JQuad asQuad(final Statement statement) {
+		return new QuadImpl(statement, salt);
```
stain: Yes, I think it would make more sense to hide the salt more and remove those static methods, perhaps have just a `RDF4JTermFactory(UUID)` constructor. Then the factory is not so locked into having to use UUID for salt.

_The reason to support a fixed `UUID` overall is if the user deliberately want them to be compatible - e.g. in a distributed architecture where two cloud instances are constructing named bnodes to be added to the same remote SAIL, or if you want reproducible output. (similar to fixed random seeds in simulations)._

I'll change it to just a constructor - removing the static methods will simplify the factory. 

In the [JenaRDFTermFactory](http://stain.github.io/incubator-commonsrdf/integration/org/apache/commons/rdf/jena/JenaRDFTermFactory.html) there's a different approach which I'm not sure what your views are on - the static methods there allow any `RDFTermFactory` to be passed in (it would use the `createBlankNode(String)` directly)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+	 *            RDF4J {@link Repository} to connect to.
+	 * @param includeInferred
+	 *            If true, any inferred triples are included in the graph
+	 * @param unionGraph
+	 *            If true, triples from any context is included in the graph,
+	 *            otherwise only triples in the default context
+	 *            <code>null</code>.
+	 * @return A {@link Graph} backed by the RDF4J repository.
+	 */
+	public RDF4JGraph asRDFTermGraph(Repository repository, boolean includeInferred, boolean unionGraph) {
```
stain: the constructor behind that method does:

```java
                if (unionGraph) {
                        // no context filter aka any context
                        this.contextFilter = new Resource[] { };
                } else {
                        // default context: null
                        this.contextFilter = new Resource[] { null };
                }
```

I think both of those should be quite common uses of `RDF4JGraph` - to select a particular graph you can also access it from [RDF4JDataset.getGraph(BlankNodeOrIRI)](http://stain.github.io/incubator-commonsrdf/integration/org/apache/commons/rdf/api/Dataset.html#getGraph-org.apache.commons.rdf.api.BlankNodeOrIRI-) -- but I agree on adding the `...` factory method so you can also expose the multi-context filter.

I'll add 

```java
public RDF4JGraph asRDFTermGraph(Repository repository, boolean includeInferred, BlankNodeOrIRI... contextFilter)
```

stain: Shall we call the boolean `anyContext` instead to be closer to RDF4J? This is after all dealing with an RDF4J `Repository`.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+	 * 
+	 * @param tripleLike
+	 *            A {@link Triple} or {@link Quad} to adapt
+	 * @return A corresponding {@link Statement}
+	 */
+	public Statement asStatement(TripleLike<BlankNodeOrIRI, org.apache.commons.rdf.api.IRI, RDFTerm> tripleLike) {
+		if (tripleLike instanceof RDF4JTripleLike) {
+			// Return original statement - this covers both RDF4JQuad and
+			// RDF4JTriple
+			RDF4JTripleLike rdf4jTriple = (RDF4JTripleLike) tripleLike;
```
stain: (You just found where `RDF4JTripleLike` is useful :)

Just a question of style I guess.. to avoid `()()` in

```java
((RDF4JTripleLike) tripleLike).asStatement()
```

the factory is long enough, so I'll go with your one-liner.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+	 * 
+	 * @param term
+	 *            RDFTerm to adapt to RDF4J Value
+	 * @return Adapted RDF4J {@link Value}
+	 */
+	public Value asValue(RDFTerm term) {
+		if (term == null) {
+			return null;
+		}
+		if (term instanceof RDF4JTerm) {
```
stain: Here of course the order is crucial as it needs to catch `RDF4JTerm` first.

I feel my style is less "nested". They are semantically the same as each `if` block does a `return`.

..but there's the danger that in the future someone edits an if-block and accidentally don't `return` (it would fall through to `throw new IllegalArgumentException`) - so I'll change to a more classical `if.. else if .. else if.. else` blocks.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+			String label = literal.getLexicalForm();
+			if (literal.getLanguageTag().isPresent()) {
+				String lang = literal.getLanguageTag().get();
+				return getValueFactory().createLiteral(label, lang);
+			}
+			org.eclipse.rdf4j.model.IRI dataType = (org.eclipse.rdf4j.model.IRI) asValue(literal.getDatatype());
+			return getValueFactory().createLiteral(label, dataType);
+		}
+		if (term instanceof BlankNode) {
+			// This is where it gets tricky to support round trips!
```
stain: I tested rdf4j-to-rdf4j roundtrip as part of the [integration-test branch](https://github.com/apache/incubator-commonsrdf/blob/jena-jsonld-rdf4j-integration/integration-tests/src/test/java/org/apache/commons/rdf/integrationtests/AllToAllTest.java).

I think the roundtrip could break if you pass blanknodes between:

* rdf4j
* jena
* rdf4j

as you would loose the .uniqueReference() and/or `.ntriplesString()` along the way.  I think if we want to support that scenario we would have to go deeper, e.g. have UUID-based blanknode labels and try our best to recognize them again.

(The JSONLD-Java `RDFTermFactory` [does this](https://github.com/apache/incubator-commonsrdf/blob/jsonld-java/jsonld-java/src/main/java/org/apache/commons/rdf/jsonldjava/JsonLdRDFTermFactory.java#L98) by recognizing the prefix in the unique reference)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
```diff
@@ -0,0 +1,509 @@
...
+			BlankNode blankNode = (BlankNode) term;
+			// FIXME: The uniqueReference might not be a valid BlankNode
+			// identifier..
+			// does it have to be in RDF4J?
```
stain: OK, I'll remove the `FIXME` and let you know if anything explodes :)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java
```diff
@@ -0,0 +1,197 @@
...
+		return c;
+	}
+
+	@Override
+	protected void parseSynchronusly() throws IOException {		
+		Optional<RDFFormat> formatByMimeType = getContentType().flatMap(Rio::getParserFormatForMIMEType);
+		String base = getBase().map(IRI::getIRIString).orElse(null);
+				
+		ParserConfig parserConfig = new ParserConfig();
+		// TODO: Should we need to set anything?
```
stain: Great, so I'll add `setParserConfig()` and `getParserConfig()`.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
```diff
@@ -0,0 +1,80 @@
...
+	protected RDF4JTermFactory rdf4jTermFactory;
+
+	public AbstractRepositoryGraphLike(Repository repository) {
+		this(repository, false);
+	}
+
+	public AbstractRepositoryGraphLike(Repository repository, boolean includeInferred) {
+		this.repository = repository;
+		this.includeInferred = includeInferred;
+		if (!repository.isInitialized()) {
```
stain: Not sure what's the difference from Commons RDF being an "RDF4J user" or the outer Commons RDF caller being a "RDF4J user" unless the caller is running `repository.initialize()` in a separate thread at the same time as passing it to RDF4JTermFactory (which sounds like asking for trouble).

If I remove this I think we can no longer do `repository.shutdown()` from `Graph.close()` - e.g. you can no longer do the `new BufferedInputStream(new InputStream)` pattern:

```
Triple t;
Sail sail = new NativeStore(file);
try (Graph g = RDF4JTermFactory.asGraph(new SailRepository(sail)) {
  g.add(t);
}
```

I think this exact scenario was why we make `api.Graph` be `AutoClosable` in the first place :)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
```diff
@@ -0,0 +1,80 @@
...
+			repository.initialize();
+			shouldWeShutdown = true;
+		}
+		rdf4jTermFactory = new RDF4JTermFactory(repository.getValueFactory());
+	}
+
+	@Override
+	public void close() throws Exception {
+		if (shouldWeShutdown) {
```
stain: Perhaps this can only be done given a `handleInitAndShutdown` boolean constructor parameter.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	
+	protected abstract T asTripleLike(Statement s);
+
+	protected RepositoryConnection getRepositoryConnection() {
+		return repository.getConnection();
+	}
+
+	public Optional<Repository> asRepository() {
```
stain: getRepositoryConnection() is empty in `ModelGraphImpl` - should that instead make a new "Model" repository connection?

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
```diff
@@ -0,0 +1,80 @@
...
+		return Optional.of(repository);
+	}
+
+	@Override
+	public Optional<Model> asModel() {
+		return Optional.empty();
```
stain: As a `RDF4JGraph` can have either a `Model` or a `Repository` - or at some point in the future possibly both - isn't the best for users to see them as `Optional` and deal with it rather than look around for exceptions?

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/BlankNodeImpl.java
```diff
@@ -0,0 +1,93 @@
...
+	@Override
+	public int hashCode() {
+		if (hashCode != 0) {
+			return hashCode;
+		}
+		return hashCode = uniqueReference().hashCode();
+	}
+
+	private boolean isValidBlankNodeLabel(String id) {
+		// FIXME: Replace with a regular expression?
```
stain: Agree - that would also remove a necessary dependency on `rdf4j-rio-turtle`.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/IRIImpl.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	public int hashCode() {
+		// Same definition
+		return value.hashCode();
+	}
+
+	@Override
+	public String ntriplesString() {
+		return "<" + value.toString() +  ">";
```
stain: `>` is not a [valid IRI character](https://tools.ietf.org/html/rfc3987#section-2.2) - I don't know if RDF4J enforces this outside parsers but thought it could be used "as is".

However [org.eclipse.rdf4j.model.IRI.toString()](http://rdf4j.org/javadoc/2.0/org/eclipse/rdf4j/model/IRI.html#toString--) is not very strict in what it defines as the "string representation" - is it safe to use, or should that javadoc be improved or an equivalent of a stronger `getIRIString()` be added to RDF4J?

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
stain: So the suggestion is to rather document `UnsupportedOperationException` on both - and eventually get to always supporting both methods? (In which case `Optional` would be silly)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
stain: The [integration test](https://github.com/apache/incubator-commonsrdf/blob/jena-jsonld-rdf4j-integration/integration-tests/src/test/java/org/apache/commons/rdf/integrationtests/AllToAllTest.java) try to add foreign blank nodes and then check that they are in `contains()` - which seem to work in all combinations (e.g. jenafactory to rdf4j graph and vice versa).

(that  did however uncover a bug in `simple.GraphImpl.contains(Triple)` in this regard)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
stain: Right.. so I shall I remove that edge case in `ModelGraphImpl` or could there be (in the future) a Model that is backed by a massive Repository?

BTW - the repository tries to ask `conn.size()` when it can:

https://github.com/apache/incubator-commonsrdf/blob/rdf4j/rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java#L86

https://github.com/apache/incubator-commonsrdf/blob/rdf4j/rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java#L94

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
stain: I had hoped it would be OK to let `Model` not deal with complex cases - perhaps I misunderstood RDF4J in thinking it is mainly a "triple store" rather than a "quad store" as it does not seem to support many context-related operations.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
stain: How can we support having the transaction on the "outside" of the `Graph` / `Dataset`..?  ADd some kind of `.begin()` and `.end()` method?

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
stain: I tried changing `asModel()` and `asGraph()` to `throw new UnsupportedOperationException()` - but it meant already in the RDF4JParser it breaks:

```java
			RDF4JDataset dataset = (RDF4JDataset) getTargetDataset().get();
			if (dataset.asRepository().isPresent()) {
				return new RDFInserter(dataset.asRepository().get().getConnection());
			}
			if (dataset.asModel().isPresent()) {
				Model model = dataset.asModel().get();
				return new AddToModel(model);
			}
```
which would then turn into a double [coding by exception](https://en.wikipedia.org/wiki/Coding_by_exception) anti-pattern with a `return` from inside `catch` .. 

So I'm inclined to keep these as `Optional`.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraph.java
stain: Should `RDF4JGraph.getContextFilter()`  (or what we call it) return RDF4J `Resource` or Commons RDF `BlankNodeOrIRI` instances?  (`Resource` is not exposed anywhere else in `RDF4JGraph`)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JIRI.java
stain: I removed the null-overrides and instead clarified the specialized generic parameter in the interface javadoc.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: Agreed it can't be null. I've changed it to a `Set<RDF4JBlankNodeOrIRI>`:

```java
public interface RDF4JGraph extends Graph, RDF4JGraphLike<Triple> {
  public Set<RDF4JBlankNodeOrIRI<Resource>> getContextMask();
}
```

Just a `Set<BlankNodeOrIRI>` would look neater, but then it's trickier to get to the `Resource`.

Not sure if a convenience method for exposing (or converting) to `Resource[]` would be good, I only needed it [here in RDF4JParser](https://github.com/apache/incubator-commonsrdf/blob/cd368d5889593dfe21b13123258b1145234a4687/rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java#L186)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: Perhaps some kind of `ThreadLocal` or transaction argument..? 

Or ask for a `TransactionGraph` that only does `.commit()` on `.close()`? 

Is a `RepositoryConnection` thread safe, or would I need `ThreadLocal`s?

I think it might be weird if the `Graph` out of the box didn't autocommit as the other implementations do - so optimizing for faster should be something you do explicitly (Commons RDF does not know how you manage your threads)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: I agree, I'll reduce the "constructor space".

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
stain: The Stream is consumed outside and is closable.. shall we just add this to the javadoc of `RDF4JGraph.stream()` that use of the stream must be closed (e.g. try-with-resources)?  (Note that this is not really true in `ModelGraph`)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
stain: I'm passing on a stream from RDF4J's `Iterations` class - so I assume its documentation is correct. But I'll add javadoc to describe the stream should be closed.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/BlankNodeImpl.java
```diff
@@ -0,0 +1,93 @@
...
+		}
+		return hashCode = uniqueReference().hashCode();
+	}
+
+	private boolean isValidBlankNodeLabel(String id) {
+		// FIXME: Replace with a regular expression?			
+		if (id.isEmpty()) { 
+			return false;
+		}
+		if (! TurtleUtil.isBLANK_NODE_LABEL_StartChar(id.codePointAt(0)))  {
```
ansell: All of the Rio parser modules other than the main package in the rio-api module are effectively implementations, even though their package names don't indicate as much. Accessing RDFParser and RDFWriter implementations is best done indirectly using ServiceLoader, via the "Rio.getParser..." methods. However, actually putting those packages down on paper as officially not part of the public API is a task that I brought up before the initial RDF4J releases but was postponed. Hence, classes like TurtleUtil are still presumed to be fair game for reuse at this point.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/IRIImpl.java
ansell: RDFWriter implementations should always check IRI's for which characters they need to escape, however, this area is likely to have a lot of bugs, as it is principally fuzz tested given the scope (Unicode) cannot be thoroughly unit tested. the IRI.toString method isn't likely to be used directly, and in general RDF4J is fairly relaxed about what it supports (via RDFParserConfig settings) because some users prefer to be able to transport with irregular data rather than fail early.

My presumption here was that ntriplesString was designed to be more strict about what it emitted compared to the Object.toString override because it followed the rules of a particular concrete format, RDF-1.1 N-Triples.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
ansell: That is great news that the integration tests are succeeding across implementations!

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
ansell: It is probably useful to leave it here. It is common knowledge that in general `Collection.size()` doesn't have `O(1)`, so this is fine for those corner cases.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java
```diff
@@ -0,0 +1,137 @@
...
+				(org.eclipse.rdf4j.model.IRI)rdf4jTermFactory.asValue(predicate), 
+				rdf4jTermFactory.asValue(object)).parallelStream()
+			.map(rdf4jTermFactory::asTriple);
+	}
+	
+	@Override
+	public Set<RDF4JBlankNodeOrIRI<Resource>> getContextMask() {
+		// ModelGraph always do the unionGraph
+		return Collections.emptySet();
+		// TODO: Should we support contextMask like in RepositoryGraphImpl?
```
ansell: Yes, Model's always contain Quads in RDF4J, so would definitely be useful to operate on them the same as Repository.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
ansell: I am not sure how to do it efficiently without explicit begin/end in the Graph/Dataset model. However, as it stands, this method will not be useful for any but the most trivial cases, given the potential overhead for commit each time so it needs work to get into circulation.

If you added begin/end then you could also optimise to reduce the number of calls to getRepositoryConnection that may be resource or time intensive depending on the Repository implementation.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
```diff
@@ -0,0 +1,172 @@
...
+	}
+	
+	@Override
+	protected RDF4JTriple asTripleLike(Statement statement) {
+		return rdf4jTermFactory.asTriple(statement);
+	}
+
+	@Override
+	public Set<RDF4JBlankNodeOrIRI<Resource>> getContextMask() {		
+		Set<RDF4JBlankNodeOrIRI<Resource>> mask = new HashSet<>();
```
ansell: Not sure how often this method will be used, but this set should not become very large in any normal situations, compared to the rest of the data, so this could be prederived in the constructor and stored in an instance variable next to contextMask.

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/NativeStoreGraphTest.java
```diff
@@ -0,0 +1,66 @@
...
+		}
+		Sail sail = new NativeStore(tempDir.toFile());
+		SailRepository repository = new SailRepository(sail);
+		repository.initialize();
+		return repository;
+	}
+	
+	@Override
+	public RDFTermFactory createFactory() {
+		// FIXME: This valueFactory will be backed by the wrong temporary folder!
```
ansell: You could use the following code to ensure that JUnit creates and cleans up a unique temporary directory for each test method:

```
@Rule 
// Note, this has to be public so JUnit can see it
public TemporaryFolder tempDir = new TemporaryFolder();

private Path testDir;

@Before
public void setUp() {
    testDir = tempDir.newFolder("test-commonsrdf-rdf4j").toPath();
}
```

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/Rdf4JRDFTermFactoryTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class Rdf4JRDFTermFactoryTest extends AbstractRDFTermFactoryTest {
+
+	@Override
+	public RDFTermFactory createFactory() {
+		return new RDF4JTermFactory();
+	}
+	
+	@Override
+	public void testInvalidLiteralLang() throws Exception {
+		Assume.assumeTrue("Sesame doesn't check Lang strings",false);
```
ansell: You can setup individual RDFParsers to check for invalid language strings using ParserConfig, but no, ValueFactory doesn't currently check for invalid language strings per its lenience in general for different types of data.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
stain: I've changed the constructors to not shutdown by default.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryDatasetImpl.java
stain: No, I was wrong.. the stream does close the `RepositoryResult`, but the repository result does not therefore close the connection (which makes sense). 

This cause the unit test to be slow shutting down (it takes 20 seconds before it times out on open connections)

I added `stream.onClose(conn::close)` which closed it correctly given regular try..with blocks outside.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: I unpublished all constructors and added an `InternalRDF4JFactory` with caveats in its javadoc.

(It needs to be public to cross the package barrier to `.impl`)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: Thanks - now fixed (assuming the stream consumer calls `.close()` or uses a `try..each` loop).

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/NativeStoreGraphTest.java
stain: Thanks, does that also work on static fields? I was trying to share the `Repository` to avoid excessive shutdown timeouts - but I guess with the fixed .stream() there might be less of those.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
stain: `Triple` does inherit from `TripleLike<BlankNodeOrIRI, IRI, RDFTerm>` - that's the whole point :)

The only thing `RDFTripleLike` adds compared to `Triple` is the `asStatement()` method - which is probably useful for integration purposes.

But I think we could add a `GeneralizedTriple` interface to avoid exposure of `TripleLike` generics - but perhaps only to the jena module for now. JSONLD-Java should in theory also support that (as JSON-LD does) - at least having blank nodes for properties. 

But I think that it can be hard - at least at this point - to say what is the "common generalized triple" at the API module.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JDataset.java
stain: I think it would be possible to expand `ModelGraphImpl` to also support contexts and even a `ModelDataset` - so I'll add that to Jira.  

I think it makes sense for now that it keeps the current behaviour which uses an empty varargs array underneath (aka the "union graph" or "disregarding contexts") - the first goal should be to have "equivalent" integration.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraph.java
stain: (Solution: `Set<BlankNodeOrIRI> getContextMask()`)

not sure which one you like least..


```
Set<RDF4JBlankNodeOrIRI<Resource>>
```

or
```
Set<BlankNodeOrIRI>
```

Sometimes with Generics you just can't win.. (The problem for RDF4JBlankNodeOrIRI is that it's subtyped by both RDF4JBlankNode and RDF4JIRI and therefor needs itself to be generic to not be in "conflict" with RDF4JTerm)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JQuad.java
stain: Good that RDF4J goes "quad first"!

That would simplify a lot both here and in jsonld-java integration, I agree. However `Triple` is where we started and also a prime type in Jena. RDF 1.1 does not really define a 'quad' as much as it defines a "triple". Also many non-expert users are likely to just want to stay at the RDF graph level and could be confused if their 'contains' call fails.

And so I think we should still support triples and graphs for the RDF4J integration.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
stain: I've removed all except `asValue(Node, UUID)` in the factory.  This is needed from `Quad` and `Triple` as they (at least now) generate on demand their RDFTerm instances of getSubject() etc.

Not sure why `java.util.UUID` is considered brittle, it's immutable. :-)  I think now with the constructors hidden it's minimized how to 'do the wrong thing'. I added javadoc for the RDF4JTermFactory constructor

If `Triple` and `Quad` instead keep the `RDF4JTermFactory` it means they also hold a reference to the `ValueFactory`  (which they don't need) - remember a Triple or Quad is what could make itself way into "other" code and so I wanted them to have as little references as possible. 

A `java.util.UUID` is quite a small object (final class, two longs) - and if for some reason needed, the code could even be optimized to keep those longs directly in the QuadImpl/TripleImpl. Or they could cache/pregenerate their getSubject/getPredicate/getObject() and not need any of the references-- it's hard to tell at this stage what is the best optimization.

ansell: Immutable objects can be brittle if they are hard to manage from a users perspective. They would need to manage the UUID and the implementation of RDFTermFactory to use it with, as they need the particular factory implementation to create objects that have the asValue/etc. methods they need.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
afs: (not sure where to comment)

"Generalised RDF" is actually a concept in RDF 1.1. The generalized RDF 1.1 came about because JSON-LD as bnodes for graph names ... and the perma-thread on literals-as-subjects.  So the WG put in the terminology for all of that.

RDF 1.1 (Generalized) does not have "quad", just RDF Dataset, Graphs, Triples and RDF Terms.  Quads are reasonably natural in that a dataset is a set of triples (default graph) and an set of quads (named graphs).

Commons RDF can either a common subset of all it integrates or try to espouse it's own take.  At the moment, it seems to me, getting it users matters.  Going too much into the details, and instead being almost a bit loose in the interests of time, could be better way to go.

As a usage : a SHACL Core engine would be a good target. It is an algorithm that can be used with various systems.  SHACL Core does not use SPARQL; some definitions in SHACL Core give both a textual and SPARQL version.

stain:
In order to close the `RepositoryConnection` (and `RepositoryResult`) I added an interface

``` java
public interface ClosableIterable<T> extends java.util.Iterable<T>, AutoCloseable { }
```

which is not very different from RDF4J's `org.eclipse.rdf4j.common.iteration.CloseableIteration` and `CloseableIterationIterator`; but can be used like this:

```java
RDF4JGraph graph; // ..
try (ClosableIterable<Triple> triples : graph.iterate()) {
   for (Triple t : triples) {
       return t; // OK to terminate for-loop early
   }
}
```

It will then close both the `RepositoryResult` and `RepositoryConnection`.

It also closes on the last `.hasNext()` - but that is for convenience only for interoperability with `Graph.iterate()` interface, and would not happen if you break out of the for loop return early (as above).

The only other alternatives to provide a somewhat transparent .iterate() with RDF4J (without the caller having to do say `.begin()` and `.end()`) I can think of would be either a `.finalize()` in the Iterator (but you won't know if/when it will run) and to consume the whole of the iteration in memory and expose a normal java.util.List iterator (which would consume lots of memory and fall over after 2G entries).

Of course interoperability is already somewhat broken with my proposal here as the other `Graph.iterate()` are not (now) ClosableIterable and thus can't be used with the try-with-resources block like that.

I updated `org.apache.commons.rdf.api.AbstractGraphTest` to avoid the RDF4J tests taking a long time (20s timeout pr test if the repository has unclosed connections). For the `.stream()` this was easy (`Stream` implements `AutoClosable`), but for the `Iterable` I had to do:

```java
private void closeIterable(Iterable<Triple> iterate) throws Exception {
        if (iterate instanceof AutoCloseable) {
        	((AutoCloseable) iterate).close();
        }
}
```

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/RepositoryGraphImpl.java
stain: Agree in principle - however I think that would be premature optimisation, `getContextMask()` is more of an "inspection hatch" method that probably won't be called much, so I don't think it needs to be stored in memory.

(That kind of caching might be much more beneficial for `org.apache.commons.rdf.rdf4j.impl.TripleImpl`)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java
```diff
@@ -0,0 +1,206 @@
...
+				throw new IOException("Can't handle source URL: " + getSourceIri().get(), ex);
+			}			
+		}
+		// must be getSourceInputStream then, this is guaranteed by super.checkSource(); 		
+		loader.load(getSourceInputStream().get(), base, formatByMimeType.orElse(null), rdfHandler);
+	}
+
+	protected RDFHandler makeRDFHandler() {
+
+		// TODO: Can we join the below DF4JDataset and RDF4JGraph cases
```
stain: I think we need to bare careful about 'carrying over' the context in statements to be added to  a `Graph` - as it could be a graph returned from `Dataset.getGraph(graphname)` - common sense would say that adding to that graph would add the triples ONLY to the `graphname` context.

However adding quads to a "union graph" (any context) could alternatively be preserving context - the current javadoc says that triples added will be added to the default context. I can remove that clause leaving that undefined; so in the future we can preserve contexts in statements added to a union graph.

In all other settings I think adding statements to a `RDF4JGraph` should respect their `contextMask` instead.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java
```diff
@@ -0,0 +1,206 @@
...
+		RDF4JParser c = prepareForParsing();
+		// Ensure we have an RDF4JTermFactory for conversion.
+		// We'll make a new one if user has provided a non-RDF4J factory
+		c.rdf4jTermFactory = (RDF4JTermFactory) getRdfTermFactory().filter(RDF4JTermFactory.class::isInstance)
+				.orElseGet(c::createRDFTermFactory);
+		return c;
+	}
+
+	@Override
+	protected void parseSynchronusly() throws IOException {
```
stain: Ah - brilliant! I was running short of use cases for an asynchronous parsing implementation!  Yes, that makes perfect sense, particularly with HTTP resources. I'll track this in Jira for the next version.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
ansell: How about adding the following interfaces to Commons RDF API, rather than experimenting inside of the integration modules that is unhelpful in the long term for reuse of similar patterns. (Note that the appropriate generics type declarations such as "`T extends RDFTerm`" have not been added to the signatures below but they would be necessary in practice):

* `GeneralisedTriple<RDFTerm, RDFTerm, RDFTerm>` 
* `GeneralisedQuad<BlankNodeOrIRI, RDFTerm, RDFTerm, RDFTerm> implements GeneralisedTriple<RDFTerm, RDFTerm, RDFTerm>` 
* `Triple implements GeneralisedTriple<BlankNodeOrIRI, IRI, RDFTerm>`
* `Quad implements Triple, GeneralisedQuad<BlankNodeOrIRI, BlankNodeOrIRI, IRI, RDFTerm>`
* `GeneralisedGraph<GeneralisedTriple, RDFTerm, RDFTerm, RDFTerm>` (need the copied generic types for methods that don't just accept GeneralisedTriple)
* `GeneralisedDataset<GeneralisedQuad, BlankNodeOrIRI, RDFTerm, RDFTerm, RDFTerm>`
* `Graph implements GeneralisedGraph<Triple, BlankNodeOrIRI, IRI, RDFTerm>`
* `Dataset implements GeneralisedDataset<Quad, BlankNodeOrIRI, BlankNodeOrIRI, IRI, RDFTerm>`

The key is that the commonly reused interfaces `Triple`/`Quad`/`Graph`/`Dataset` now do not have the confusing "Like" suffix, and they have no generics arguments to remove an extra barrier to reuse. All implementations of Triple will be specialised subsets of GeneralisedTriple, so it doesn't seem to be an issue to uptake to represent the APIs this way.

Inheriting `Quad` from `Triple` and other similar instances isn't a design issue IMO, as correct code should be written to check `instanceof Quad` before it checks `instanceof Triple` if the user wants to preserve the dataset reference.

Even though Quad is not defined in terms of RDF-1.1, there will be no simple way to integrate with RDF4J without some version of it. It is also a commonly used term in the community, even if it didn't make it into the RDF-1.1 specification so it doesn't introduce new issues for understandability.

Implementing a Triple-level integration for RDF4J would not be a useful pursuit IMO given the background of the RDF4J Statement interface always having had a getContext method. I would prefer if this integration went just for `Quad<->Statement` translation in its initial versions to get it completed sooner rather than later.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
```diff
@@ -0,0 +1,63 @@
...
+	 * 
+	 * @return The corresponding RDF4J Model.
+	 */
+	public Optional<Model> asModel();
+
+	/**
+	 * Return the corresponding RDF4J {@link Repository}, if present.
+	 * <p>
+	 * The return value is {@link Optional#isPresent()} if this is backed by a
+	 * Repository.
```
ansell: The return value isn't `Optional.isPresent()`. It will return `true` for `Optional.isPresent()` and `Optional.get()` will then need to be called to get the underlying `Repository`. Easier to refer to `Optional.empty()` as the return value when it is not backed by a `Repository`.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JIRI.java
ansell: See my comment about removing the generics parameter from RDF4JTerm so it would no longer be a null override, but would be using typical Java return value extension to redefine the return value. That would simplify understanding here IMO, with "asValue" returning "IRI", but having no reason to have to pass that on to the super interface, as the heirarchy already matches the Value->IRI heirarchy from RDF4J so should work smoothly.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JLiteral.java
ansell: I actaully prefer this method now. If generics are removed from RDF4JTerm it would no longer be a null override and would still return a Literal as expected.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java
ansell: Feel free to implement a `Graph`/`Triple` lifecycle independent of the `Dataset`/`Quad` lifecycle, but it doesn't seem like it has any particular benefit or attraction for RDF4J users who have always had `getContext()` as a method on `Statement`.

ansell:
Adding CloseableIterable just to RDF4J integration module will not help interoperability, as code will actually need to be rewritten to switch to another implementation due to the restrictions on how try-with-resources accepts inputs. If you want that part to work transparently you will need to implement it in the base Commons RDF API.

stain:
Thanks for the great and thorough review, @ansell! I think this helped get a better shape of the interfaces.

I used some of the feedback also on the `jena` module, which now also reduced most of the `UUID` exposure.

RDF4J is interesting because of the difference in semantics around datasets - I think one value of Commons RDF is that we give a (hopefully) unified view of semantics between the implementations - and then of course users can "upgrade" to native RDF4J if they want other semantics or more functionalities.

I'll raise in Jira the remaining issues I am afraid I don't have time to look at right now, and go ahead to merge this to `master`.

Once merged we can see if we can normalize the converter method names between the factories.

Then add the integration tests, and roll for a `0.3.0-incubating` release.

Cheers!

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JGraphLike.java
stain: Tracked as https://issues.apache.org/jira/browse/COMMONSRDF-42 - let's discuss it further there.

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JLiteral.java
stain: Done :-)

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/BlankNodeImpl.java
stain: Raised as https://issues.apache.org/jira/browse/COMMONSRDF-43

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4JTermFactory.java
stain: Booleans changed to `Options...` enum.

As a side-effect this makes it possible to deliberately ask for the initAndShutdown (e.g. used by `createDataset()`

rdf4j/pom.xml
stain: They should not need the rest of RDF4J, but can use the normal RDF4J all-in dependency if they like ; it would be compatible. I think this can go into docs.

Adding everything seemed to add multiple external libraries that might not be needed or might be in conflict with something else in a Commons RDF user's project (which after all might be doing integration).

rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/AbstractRepositoryGraphLike.java
stain: I added to the javadoc that at least one of them will not be empty.

PR #25, opened 2016-10-23
COMMONSRDF-46 Rename RDFTermFactory to RDF

stain:
[COMMONSRDF-46](https://issues.apache.org/jira/browse/COMMONSRDF-46)

This keeps a `@Deprecated` skeleton `RDFTermFactory` with the 0.2.0 signatures, adds a newer `RDF` interface which also includes the new `createDataset()` and implementation-specific methods.

The `RDF` implementation also change names to match, e.g. `RDF4J` instead of `RDF4JFactory`, `JenaRDF` instead of `JenaFactory`.

This pull requests includes suggested [changes to the user guide](https://github.com/apache/incubator-commonsrdf/blob/COMMONSRDF-46/src/site/markdown/userguide.md#creating-commons-rdf-instances)

See also [suggested javadoc](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-46/org/apache/commons/rdf/api/RDF.html) and children.


stain:
Any comments, @wikier  @ansell  @afs @ajs6f  @lewismc  ?

I'll merge this on Friday unless I hear otherwise :-)

wikier:
Sound good to me. `TermFactory` was legacy from Sesame. So it's good to have new fresh ideas coming-in that could help the API usage.

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
```diff
@@ -200,7 +200,7 @@
 	}
 
 
-	private Optional<RDFTermFactory> rdfTermFactory = Optional.empty();
```
ansell: The use of optional here as a field type makes it impossible to serialise. Need to have the raw values stored in fields if you want to support serialisation in the future, which should otherwise be possible.

Even if it sounds strange, I have had requests about this in the past so people do use serialisation functionality if it is available.

src/site/markdown/userguide.md
```diff
@@ -217,7 +222,7 @@ objects as part of data parsing, storage lookup and queries.
 the super-interface for instances that can be used as subject, predicate and
 object of a [Triple](apidocs/org/apache/commons/rdf/api/Triple.html).
 
-The RDF term interfaces are arranged in this class hierarchy:
+The RDF term interfaces are arranged in this ty[e hierarchy:
```
ansell: ty[e -> should be "type"

api/src/main/java/org/apache/commons/rdf/api/RDF.java
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * Create a new graph.
+     *
+     * It is undefined if the graph will be persisted by any underlying storage
+     * mechanism.
+     *
+     * @return A new Graph
+     */
+	 public Graph createGraph();
```
stain: Thanks, I'll do a separate commit to fix indentation to all spaces, which I think is what Commons prefer. (That got quite a large diff, so I won't pollute this PR with it)

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -31,7 +31,7 @@
  * and <a href="https://en.wikipedia.org/wiki/TriX_%28syntax%29">TriX</a>.
  * 
  * @see <a href="https://www.w3.org/TR/rdf11-primer/#section-graph-syntax">RDF 1.1 Primer</a>
- *
+ * @see org.apache.commons.rdf.experimental.RDFParser
  */
 public enum RDFSyntax {
```
stain: Raised in [COMMONSRDF-47](https://issues.apache.org/jira/browse/COMMONSRDF-47)

It was made an enum as it is documented as the RDF 1.1 serializations, which is constant. But I agree an interface would be better - that could also simplify the `RDFParser` interface which now needs to have a `String` as a fallback.

I think we can change it so it is somewhat compile-compatible with 0.2.0  (e.g. `RDFSyntax x = RDFSyntax.TURTLE`) even if not binary compatible (enums are kept as numbers) - it is 
mainly useful with the Parser interface which we are adding now in upcoming 0.3.0.

api/src/main/java/org/apache/commons/rdf/api/RDFTermFactory.java
```diff
@@ -17,262 +17,61 @@
...
-     * Create a new blank node.
-     * <p>
-     * The returned blank node MUST NOT be equal to any existing
-     * {@link BlankNode} instances according to {@link BlankNode#equals(Object)}.
-     *
-     * @return A new, unique {@link BlankNode}
-     * @throws UnsupportedOperationException
-     *             If the operation is not supported.
-     */
     default BlankNode createBlankNode() throws UnsupportedOperationException {
```
stain: Yes, I removed the notion of `UnsupporetedOperationException` from `RDF` as all our bundled implementations are complete and callers should not be too worried about that. 

The Javadoc now say that any partial `RDF` implementations should be _clearly documented as such_:

>  * A <em>partial RDF implementation</em> 
>  * should be clearly documented as such, and may throw 
>  * {@link UnsupportedOperationException} where applicable, e.g. if it 
>  * does not support creating {@link Dataset}s or {@link Quad}s. 

Is this sufficient? E.g. an `RDF` instance within OWLAPI probably would not know how to make a `Graph` or `Dataset` - but would be a bit more exotic factory than the ones bundled within Commons RDF.

api/src/test/java/org/apache/commons/rdf/api/AbstractRDFTermFactoryTest.java
```diff
@@ -208,8 +172,8 @@ public void testCreateIRIRelative() throws Exception {
         // be possible to resolve to an absolute IRI)
         try {
             factory.createIRI("../relative");
-        } catch (UnsupportedOperationException | IllegalArgumentException ex) {
-            Assume.assumeNoException(ex);
+        } catch (IllegalArgumentException ex) {
+            Assume.assumeNoException("Relative IRIs not supported - ignore this test", ex);
```
stain: Agreed - perhaps a factory would support relative IRIs but would make them absolute - e.g. the factory could have its own base URI (as would be the case when parsing). Removed test.

jena/src/test/java/org/apache/commons/rdf/jena/TestJenaGraphToCommonsRDFGraph.java
```diff
@@ -33,7 +32,7 @@
 import org.apache.commons.rdf.api.RDFTerm;
 import org.apache.commons.rdf.api.Triple;
 import org.apache.commons.rdf.simple.Types;
-import org.apache.jena.riot.Lang;
+import org.apache.jena.riot.Lang ;
```
stain: Well spotted! I think it came from @afs's original code  - I have removed the spaces.

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/NativeStoreGraphTest.java
```diff
@@ -49,17 +51,23 @@
  */
 public class NativeStoreGraphTest extends AbstractGraphTest {
 
-	public final class NativeStoreFactory implements RDFTermFactory {
+	public final class NativeStoreRDF implements RDF {
```
stain: Just to satisfy `AbstractGraphTest` which blindly calls `createGraph()` on the `RDF` factory. The functionality is also there in `RDF4J` using `asGraph`.   Perhaps a richer `NativeStore` factory could be made that hides the `ValueFactory` plumbing and can be created with a folder `Path`?  

Unclear if such a factory would then reuse the same connection for each `createGraph()` and not support `createDataset()` (more than once), or if it would just return the populated Dataset, or if each call would make a new connection in a new folder.. 

Let's discuss it in [COMMONSRDF-48](https://issues.apache.org/jira/browse/COMMONSRDF-48) as I think that would be generally useful to have a shortcut. (and as well for Jena backed by TDB)

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/NativeStoreGraphTest.java
```diff
@@ -49,17 +51,23 @@
...
 		@Override
 		public RDF4JGraph createGraph() {
 			// We re-use the repository connection, but use a different context every time
 			Set<RDF4JBlankNode> context = Collections.singleton(rdf4jFactory.createBlankNode());
 			return rdf4jFactory.asGraph(getRepository(), context);
 		}
-
+		@Override
+		public Dataset createDataset() {
+			throw new UnsupportedOperationException("Can't create more than one Dataset in this test");
```
stain: I tried to do that, yes. The problem then was that you couldn't easily see what caused the 30s delay if something was not properly closed. Now that everything is closed properly within the tests it doesn't take many seconds to run all of the tests.

rdf4j/src/test/java/org/apache/commons/rdf/rdf4j/Rdf4JRDFTermFactoryTest.java
```diff
@@ -18,19 +18,19 @@
...
-	public RDFTermFactory createFactory() {
-		return new RDF4JFactory();
+	public RDF createFactory() {
+		return new RDF4J();
 	}
 	
 	@Override
 	public void testInvalidLiteralLang() throws Exception {
-		Assume.assumeTrue("Sesame doesn't check Lang strings",false);
+		Assume.assumeTrue("RDD4J doesn't check Lang strings",false);
```
stain: Fixed. Also renamed `*RDFTermFactoryTest` to `*RDFTest`

I changed to `@Ignore` in `JsonLdRdfTest`, but I can't use @Ignore here as the parent test method has `@Test(expected = IllegalArgumentException.class)`  and somehow the `@Ignore` is not honoured - at least in Eclipse.

Using `Assume.assumeTrue` is still recognized as "test skipped".

api/src/main/java/org/apache/commons/rdf/api/BlankNodeOrIRI.java
```diff
@@ -19,7 +19,7 @@
 
 /**
  * This interface represents the {@link RDFTerm}s that may be used in the
- * subject position of an RDF-1.1 {@link Triple}, including {@link BlankNode}
+ * subject position of an RDF-1.1 {@link Triple}, that is {@link BlankNode}
```
stain: "either" is stronger - meaning that we say they are distinct. (Which I think they are)

if you agree we can then also say in `RDFTerm` that it also must be a `IRI`, `BlankNode` or `Literal`.  Perhaps using _SHOULD_ so that evil people can still make their `SparqlVariable` if they so wish - knowing well that this would break various stuff.

In `RDFTerm` I added:

 * Instances of RDFTerm SHOULD be an {@link IRI}, {@link BlankNode} or
 * {@link Literal}.

and in `BlankNodeOrIRI`:

>  * Instances of BlankNodeOrIRI SHOULD be a {@link BlankNode} or an {@link IRI}.

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
stain: Yes, I think we should review this post 0.3.0 as we de-experimentalize `AbstractRDFParser` and make equivalent `Writer` - there's nothing else preventing it from being serializable - the `RDF` factory itself must also be serializable though.

Raised as
[COMMONSRDF-49](https://issues.apache.org/jira/browse/COMMONSRDF-49)

stain:
Thanks for the review, @ansell @wikier @ajs6f !

Merging now and then adding the reformat to spaces. 

@wikier and @ajs6f  - would you agree with Peter on being brave and remove all of `RDFTermFactory`  rather than `@deprecated`? It's compatible with [Semantic Versioning for 0.x series](http://semver.org/spec/v2.0.0.html#spec-item-4) - but I thought it would be better to keep it for one round just to warn people off given that we introduced it in 0.2.0 - they need to switch anyway to use the Jena or Sesame integrations.

I don't see a problem with keeping `@since` even if we are pre-1.0.0. It's still useful for anyone looking at the Javadoc.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
ajs6f: I wouldn't feel beholden to be all that compatible with the previous release, since this is all < release 1.0.

api/src/main/java/org/apache/commons/rdf/api/RDFTermFactory.java
ajs6f: Maybe we could have a specific marker interface for that `PartialRDF` < `RDF`? Or a test method `default boolean isComplete() {return false;}`?

ajs6f:
Hm. It _would_ seem cleaner. I wouldn't feel it as a terrible downstream burden; after all, this is < release 1.0.

PR #26, opened 2016-10-27
COMMONSRDF-7 RDFTerm/Triple/Quad are immutable

stain:
[COMMONSRDF-7](https://issues.apache.org/jira/browse/COMMONSRDF-7)

Added paragraphs about immutability to `RDFTerm`, `Triple` and `Quad` (but not to `TripleLike` or `QuadLike`)

also clarify `RDFTerm.equals()` and `RDFTerm.hashCode()`
by delegation to its specializations IRI/BlankNode/Literal

Suggested javadoc: 

* [RDFTerm](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-7/org/apache/commons/rdf/api/RDFTerm.html)
* [Triple](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-7/org/apache/commons/rdf/api/Triple.html)
* [Quad](http://stain.github.io/incubator-commonsrdf/COMMONSRDF-7/org/apache/commons/rdf/api/Quad.html)


stain:
This clarifies immutability of an `RDFTerm` (as it is already implemented) and that it can be safely used in say a `Set`.

I added `.equals()` and `.hashCode()` to `RDFTerm` which delegate to `IRI`, `BlankNode` and `Literal`.

One questions is if `RDFTerm` class should be restrictive about its children being either an `IRI`, `BlankNode` or `Literal` - which would not allow like `Variable` we tried in Jena - or multiple inheritance.

stain:
@ajs6f - would you be willing to help review..?

api/src/main/java/org/apache/commons/rdf/api/Quad.java
```diff
@@ -17,16 +17,30 @@
...
  * A Quad is a statement in a
  * <a href= "http://www.w3.org/TR/rdf11-concepts/#section-dataset" >RDF-1.1
  * Dataset</a>, as defined by
- * <a href= "https://www.w3.org/TR/2014/NOTE-rdf11-datasets-20140225/" >RDF-1.1
+ * <a href= "https://www.w3.org/TR/2014/NOTE-rdf11-datasets-20140225/#quad-semantics" >RDF-1.1
  * Concepts and Abstract Syntax</a>, a W3C Working Group Note published on 25
  * February 2014.
+ * <p>
+ * A {@link Quad} object in Commons RDF is considered <em>immutable</em>, that
+ * is, over it's life time it will have consistent behaviour for its
```
ajs6f: typo: `its` in both places

api/src/main/java/org/apache/commons/rdf/api/Quad.java
```diff
@@ -17,16 +17,30 @@
...
+ * {@link #equals(Object)} and {@link #hashCode()}, and the instances returned
+ * from {@link #getGraphName()}, {@link #getSubject()}, {@link #getPredicate()}
+ * and {@link #getObject()} will have consistent {@link Object#equals(Object)}
```
ajs6f: Do we need to talk about `hashCode` at all? Don't the definitions of those methods prevent a situation where we have `equals`-consistency but not `hashCode`-consistency?

api/src/main/java/org/apache/commons/rdf/api/Quad.java
```diff
@@ -17,16 +17,30 @@
...
+ * behaviour.
+ * <p>
+ * Thus, a {@link Quad} is thread-safe and can be safely used in collections
```
ajs6f: For thread-safety, must we not also either guarantee that any `setSubject` etc. methods are atomic or that there are none (much the likelier story)?

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
```diff
@@ -17,14 +17,32 @@
...
- * 25 February 2014.<br>
+ * Term</a>, as defined by
+ * <a href= "http://www.w3.org/TR/rdf11-concepts/" >RDF-1.1 Concepts and
+ * Abstract Syntax</a>, a W3C Recommendation published on 25 February 2014.
+ * <p>
+ * A {@link RDFTerm} represents either an {@link IRI}, a {@link BlankNode} or a
+ * {@link Literal}.
+ * <p>
+ * A {@link RDFTerm} object in Commons RDF is considered <em>immutable</em>,
+ * that is, over it's life time it will have consistent behaviour for its
```
ajs6f: typo: `its` in both places

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
```diff
@@ -17,14 +17,32 @@
...
+ * {@link #equals(Object)} and {@link #hashCode()}, and objects returned from
+ * its getter methods (e.g. {@link IRI#getIRIString()} and
+ * {@link Literal#getLanguageTag()}) will have consistent
+ * {@link #equals(Object)} behaviour.
```
ajs6f: See comment above about `equals` vs. `hashCode`.

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
```diff
@@ -17,14 +17,32 @@
...
+ * <p>
+ * Thus, an {@link RDFTerm} is thread-safe and can be safely used in collections
+ * like {@link List}, {@link Map} or {@link Set}, and a {@link RDFTerm} can be
```
ajs6f: Not sure what being used in a `Collection` has to do with thread-safety, but maybe I'm reading this wrong?

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
```diff
@@ -37,4 +55,46 @@
      */
     String ntriplesString();
 
+	/**
+	 * Check it this RDFTerm is equal to another RDFTerm.
+	 * <p>
+	 * If this object is an {@link IRI}, equality is checked using
```
ajs6f: If this must _always_ be true, sounds like an opportunity to use a `default` method here.

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
```diff
@@ -37,4 +55,46 @@
...
+    public boolean equals(Object other);
+
+	/**
+	 * Calculate a hash code for this RDFTerm.
+	 * <p>
+	 * As an {@link RDFTerm} is <em>immutable</em>, this method will always
+	 * return the same hashCode over the lifetime of this object.
+	 * <p>
+	 * This method MUST be implemented in conjunction with
+	 * {@link #equals(Object)} so that two equal RDFTerm produce the same hash
```
ajs6f: typo: `RDFTerms`

api/src/main/java/org/apache/commons/rdf/api/Triple.java
```diff
@@ -17,18 +17,32 @@
...
  * An <a href= "http://www.w3.org/TR/rdf11-concepts/#dfn-rdf-triple" >RDF-1.1
- * Triple</a>, as defined by <a href= "http://www.w3.org/TR/rdf11-concepts/"
- * >RDF-1.1 Concepts and Abstract Syntax</a>, a W3C Recommendation published on
- * 25 February 2014.<br>
+ * Triple</a>, as defined by
+ * <a href= "http://www.w3.org/TR/rdf11-concepts/" >RDF-1.1 Concepts and
+ * Abstract Syntax</a>, a W3C Recommendation published on 25 February 2014.<br>
+ * <p>
+ * A {@link Triple} object in Commons RDF is considered <em>immutable</em>, that
+ * is, over it's life time it will have consistent behaviour for its
```
ajs6f: typo: `its` in both places

ajs6f:
Generally, looks good to me. The questions I have are:

1. Do we need to talk about `equals` and `hasCode` separately?
2. Can we clarify the definition of "immutable" a bit: most people, I think, would assume that it means that e.g. `Quad::getSubject` is actually going to return the identical object, not an `equals` object. I'm not against the definition you are using here, and I can understand how it might give a useful flexibility during implementation, but it might be nice to have a higher-level note explaining it carefully.
3. I think there could be a use of `default` methods here.

api/src/main/java/org/apache/commons/rdf/api/Quad.java
stain: Documenting `Quad` as immutable would indeed prevent any `setSubject()` bean unless you made them write-once - stringently throwing `IllegalStateException` until all of them were set.

You could however make such a Java bean from `QuadLike` which don't have immutability restriction, and then have a `.build()` method or similar that return the immutable `Quad`.

api/src/main/java/org/apache/commons/rdf/api/Quad.java
stain: I removed the `hashCode` text and rather said that `Triple`, `Quad` and `RDFTerm` can be used in hashing collections like `HashSet`.

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
stain: It's not - they are both due to immutability. I split this into multiple paragraphs.

api/src/main/java/org/apache/commons/rdf/api/RDFTerm.java
stain: I don't think that would make sense in `RDFTerm` as each of them override `.equals()`? What would the `default` method do and when would it be called? 

Another problem is that you can't provide a `default` method for methods from `java.lang.Object`.

ajs6f:
Oh, my goodness-- you're quite right of course. I just spaced out about that.

PR #27, opened 2016-10-28
COMMONSRDF-47 RDFSyntax as an interface

stain:
[COMMONSRDF-47](https://issues.apache.org/jira/browse/COMMONSRDF-47)

RDFSyntax  as an interface, with OfficialRDFSyntax is the enum - which implements the interface

basically .mediaType -> .mediaType() etc

One problem is that you can't override `.equals()` on an enum - and presumably it might be useful to compare if two RDFSyntaxes are the same? 

Do we need to add our own `.equals()` and `.hashCode()` contract here as well (using `mediaType()`)  and get rid of the enum?

we will also have to tweak the japicmp config as it (rightfully) complains about `RDFSyntax` changing its signature since 0.2.0, e.g. 

```
***! MODIFIED INTERFACE (<- ENUM) : PUBLIC ABSTRACT (<- NON_ABSTRACT) NON_FINAL (<- FINAL) org.apache.commons.rdf.api.RDFSyntax  (class type modified)
        ---* REMOVED INTERFACE: java.lang.Comparable
        ---* REMOVED INTERFACE: java.io.Serializable
        ***! MODIFIED SUPERCLASS: java.lang.Object (<- java.lang.Enum)
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) RDFA_XHTML
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) NQUADS
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) RDFA_HTML
        ---! REMOVED FIELD: PUBLIC(-) FINAL(-) java.lang.String mediaType
        ---! REMOVED FIELD: PUBLIC(-) FINAL(-) boolean supportsDataset
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) RDFXML
        ---! REMOVED FIELD: PUBLIC(-) FINAL(-) java.lang.String fileExtension
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) TURTLE
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) NTRIPLES
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) TRIG
        ***! MODIFIED FIELD: PUBLIC STATIC FINAL org.apache.commons.rdf.api.RDFSyntax$OfficialRDFSyntax (<- org.apache.commons.rdf.api.RDFSyntax) JSONLD
        +++* NEW METHOD: PUBLIC(+) ABSTRACT(+) java.lang.String fileExtension()
        +++* NEW METHOD: PUBLIC(+) ABSTRACT(+) java.lang.String mediaType()
        +++* NEW METHOD: PUBLIC(+) ABSTRACT(+) java.lang.String name()
        +++* NEW METHOD: PUBLIC(+) ABSTRACT(+) boolean supportsDataset()
        +++* NEW METHOD: PUBLIC(+) ABSTRACT(+) java.lang.String title()
        ---  REMOVED METHOD: PUBLIC(-) java.lang.String toString()
        ---! REMOVED METHOD: PUBLIC(-) STATIC(-) org.apache.commons.rdf.api.RDFSyntax valueOf(java.lang.String)
        ---! REMOVED METHOD: PUBLIC(-) STATIC(-) org.apache.commons.rdf.api.RDFSyntax[] values()
```

My suggestion is to delay this till past 0.3.0, as we're not settling `RDFParser` now anyway.


api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -34,7 +152,7 @@
  *      1.1 Primer</a>
  * @see org.apache.commons.rdf.experimental.RDFParser
  */
-public enum RDFSyntax {
+  public enum OfficialRDFSyntax implements RDFSyntax {
```
ansell: There is no particular reason that "enum" needs to be used given the interface design is done for flexibility, and given the difficulties with equals and hashCode. I would suggest reimplementing the "values()" method, along with equals and hashCode, and changing this to a "public final class".

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
```diff
@@ -240,7 +240,7 @@ public T rdfTermFactory(RDF rdfTermFactory) {
     public T contentType(RDFSyntax rdfSyntax) throws IllegalArgumentException {
         AbstractRDFParser<T> c = clone();
         c.contentTypeSyntax = Optional.ofNullable(rdfSyntax);
-        c.contentType = c.contentTypeSyntax.map(syntax -> syntax.mediaType);
+        c.contentType = c.contentTypeSyntax.map(syntax -> syntax.mediaType());
         return c.asT();
```
ansell: It hurts my head thinking about this, but what is the signature for the generic type "T" designed to do. It seems like a circular definition, which doesn't really make sense but I can't think why T could have been introduced so I am stuck at seeing it as circular.

```
AbstractRDFParser<T extends AbstractRDFParser<T>>
```

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
```diff
@@ -240,7 +240,7 @@ public T rdfTermFactory(RDF rdfTermFactory) {
     public T contentType(RDFSyntax rdfSyntax) throws IllegalArgumentException {
         AbstractRDFParser<T> c = clone();
```
ansell: Is it actually necessary to clone here? The typical builder pattern is much simpler with its reliance on always having "return this" as its last line, and using mutation before the "build()" terminal method to reduce this horrendous complexity related to cloning longterm mutable objects.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -22,6 +22,124 @@
...
+     * @param mediaType
+     *            The media type to match
+     * @return If {@link Optional#isPresent()}, the {@link RDFSyntax} which has
+     *         a matching {@link RDFSyntax#mediaType}, otherwise
+     *         {@link Optional#empty()} indicating that no matching syntax was
+     *         found.
+     */
+    public static Optional<RDFSyntax> byMediaType(String mediaType) {
+        final String type = mediaType.toLowerCase(Locale.ENGLISH).split("\\s*[;,]", 2)[0];
+        return Arrays.stream(OfficialRDFSyntax.values()).filter(t -> t.mediaType().equals(type))
```
ansell: If OfficialRDFSyntax was not an enum, then you could redefine the .values() method to return Set<RDFSyntax> and remove the .map call below that would then be performing an unnecessary cast.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -22,6 +22,124 @@
...
+     * 
+     * @param mediaType
+     *            The media type to match
+     * @return If {@link Optional#isPresent()}, the {@link RDFSyntax} which has
+     *         a matching {@link RDFSyntax#mediaType}, otherwise
+     *         {@link Optional#empty()} indicating that no matching syntax was
+     *         found.
+     */
+    public static Optional<RDFSyntax> byMediaType(String mediaType) {
+        final String type = mediaType.toLowerCase(Locale.ENGLISH).split("\\s*[;,]", 2)[0];
```
ansell: Is this an official RFC/standards-based regex for splitting media-types?

acoburn: According to https://www.ietf.org/rfc/rfc2045.txt, it seems that splitting on `\\s*;` would be more accurate.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -22,6 +22,124 @@
...
+    /**
+     * The <a href="https://tools.ietf.org/html/rfc2046">IANA media type</a> for
+     * the RDF syntax.
+     * <p>
+     * The media type can be used as part of <code>Content-Type</code> and
+     * <code>Accept</code> for <em>content negotiation</em> in the
+     * <a href="https://tools.ietf.org/html/rfc7231#section-3.1.1.1">HTTP
+     * protocol</a>.
+     */
+    public String mediaType();
```
stain: No, only the official media type and extension should be used in this interface - however subtypes might add `mediaTypes()` and similar, or list several `RDFSyntax` instances (which would then not be `.equal()`).

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -22,6 +22,124 @@
...
+
+    /**
+     * The <a href="https://tools.ietf.org/html/rfc2046">IANA-registered</a>
+     * file extension.
+     * <p>
+     * The file extension includes the leading period, e.g. <code>.jsonld</code>
+     */
+    public String fileExtension();
```
stain: No :-)

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -22,6 +22,124 @@
...
+
+    /**
+     * Indicate if this RDF syntax supports
+     * <a href="https://www.w3.org/TR/rdf11-concepts/#section-dataset">RDF
+     * Datasets</a>.
+     */
+    public boolean supportsDataset();
+
+
```
stain: I think that is out of scope for now, as Commons RDF does not specify namespace preservation yet anywhere else.

Let's revisit that when we make `RDFWriter` where it might be a bit more natural than in `RDFParser` (although I know RDF4J can preserve parsed namespaces)

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
stain: I changed it to only support the media type format as @acoburn  suggests, not the content type which can list multiple types (parsing a `Content-Type` header properly requires more work!)

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
stain: Changed `OfficialRDFSyntax`  to a package-protected class `W3CRDFSyntax` and referred to its instances.

`values()` renamed to `w3cSyntaxes()` which return a `Set<RDFSyntax>`. I don't think we need to make a `HashMap` from media type/extension to `RDFSyntax` as there's only 8 instances to iterate over anyway.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
stain: I also wanted to list the `IRI` identifying the syntaxes, but then I would need a complete `IRIImpl` within `api/` .. or move `W3CRDFSyntax` to `simple/`.  (or.. awkwardly.. ask for an `RDF` instance :))

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
stain: No, but cloning means the `RDFParser` is immutable and can be reused in a thread-safe manner, e.g. if parsing many Turtle files in the same folder. So what is worse..:

* a) Creating 1000 RDFParser instances, setting all parameters every time, and remember not to reuse/share in multiple threads.
* b) Create a single base RDFParser instance with the common settings, then specialize only what is needed for each file to use in any thread at any time. (e.g. called from a within a `Stream.parallell().map()`

I think the overhead of cloning here is pretty much equivalent, e.g. `1000` instances, or `1005` instances.  

If you are parsing just a single file then I don't think the inefficiency of the cloning makes a big difference - that is unless you are making a very snappy UI or something.

Let's change this to mutable only if benchmark numbers show a problem?

simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
stain: It's just a generics trick so that subclasses of `AbstractRDFParser` can return their own type in all the setter methods rather than `RDFParser` or `AbstractRDFParser`, which would hide any additional methods they specialize with.

api/src/test/java/org/apache/commons/rdf/api/RDFSyntaxTest.java
```diff
@@ -114,8 +116,7 @@ public void name() throws Exception {
 
     @Test
     public void valueOf() throws Exception {
-        assertEquals(RDFSyntax.TURTLE, RDFSyntax.valueOf("TURTLE"));
-        // No need to test all of them, we'll trust Enum
+        assertEquals(RDFSyntax.TURTLE, RDFSyntax.byName("TURTLE"));
```
acoburn: `RDFSyntax::byName` returns an `Optional`. Perhaps you mean:

    assertEquals(RDFSyntax.TURTLE, RDFSyntax.byName("TURTLE").get());

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -178,17 +206,62 @@ private RDFSyntax(final String name, final String mediaType, final String fileEx
...
-        return Arrays.stream(RDFSyntax.values()).filter(t -> t.fileExtension.equals(ext)).findAny();
+        final String ext = fileExtension.toLowerCase(Locale.ROOT);        
+        return w3cSyntaxes().stream().filter(t -> t.fileExtension().equals(ext))
+                .findAny();
+    }
+    
+    /**
+     * Return the RDFSyntax with the specified {@link #name()}.
+     * <p>
+     * The list of syntaxes supported is at least those returned by
```
ansell: There doesn't look to be any way to extend that set from a user point of view. This wording implies there is a way.

stain: Changed to 
> This method support all syntaxes returned by {@link #w3cSyntaxes()}

There is no good way from a static class method to allow user extension; at least without having a discovery mechanism (classpath sensitive) or mutable setters/registrations (initialization sensitive) - I would argue it is out of scope for this particular method to support that, as Commons RDF only  target RDF 1.1.  

If you want we could change these lookup method to take a variable/optional list of `Iterable<RDFSyntax>` as parameters? 

(The parser/writer methods would with this PR support `RDFSyntax` from "elsewhere" - e.g. we could return supported syntaxes from each `RDF` instance)

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
```diff
@@ -91,7 +148,30 @@
...
+     * 
+     * @return
+     *      A set of the official RDF 1.1 {@link RDFSyntax}es.
+     * 
+     * @see <a href="https://www.w3.org/TR/rdf11-primer/#section-graph-syntax">RDF
+     *      1.1 Primer</a>
+     * @see org.apache.commons.rdf.experimental.RDFParser
+     */
 
+    public static Set<RDFSyntax> w3cSyntaxes() {
```
stain: `W3CRDFSyntax` is package-protected. It was moved it out followingyour advice, but I don't see any win in making it `public` as it has no additional methods (and shouldn't have).

If we make it public and remove `w3cSyntaxes()` , then there's not much more any point to have the proxy constants `RDFSyntax.TURTLE` etc. as they would also then be made public at `W3CRDFSyntax.TURTLE` etc - which - not withstanding the CAPSISSUE - is not as accessible for the 95% use cases of parsing/writing a RDF 1.1 syntax.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
stain: So my suggestion was that `mediaType()` has "The IANA media type" and similar for `fileExtension()`. Arbitrary "other" media types and extensions would have to be in a different method - perhaps the plural `mediaTypes()` and `fileExtensions()`?

There is nothing stopping a parser from supporting other media types and file extensions, but I'm not sure why they also need to be present in the `RDFSyntax` interface, which main purpose is to hard-code a syntax for reading (and maybe more so writing).

The previously proposed [RDFParser](https://github.com/apache/commons-rdf/blob/master/api/src/main/java/org/apache/commons/rdf/experimental/RDFParser.java) interface - which I hope we can discuss in a separate pull request / Jira Issue - supports  both `contentType(String contentType)` and `contentType(RDFSyntax rdfSyntax)`.

It also has for `source(IRI)`:

>     * The {@link #contentType(RDFSyntax)} or {@link #contentType(String)} MAY
>     * be set before calling {@link #parse()}, in which case that type MAY be
>     * used for content negotiation (e.g. <code>Accept</code> header in HTTP),
>     * and SHOULD be used for selecting the RDFSyntax.


while for `source(Path)` (a file) it changes to SHOULD (as file extensions are fragile)

>     * The {@link #contentType(RDFSyntax)} or {@link #contentType(String)}
>     * SHOULD be set before calling {@link #parse()}.

This hints that a parser is able to do file extension or file magic, but is not required to. Both RDF4J and Jena implementations have different mechanisms for this, e.g. [RDF4JParser](https://github.com/apache/commons-rdf/blob/master/rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/experimental/RDF4JParser.java#L140) falls back to `Rio.getParserFormatForFileName()` based on path filename if no content type is specified - which would use RDF4J's extended list of file extensions and formats.

Obviously this would be easy to implement consistently for our hard-coded w3c-standardized file extensions (and we could document that), but much harder for arbitrary "other" extensions.

api/src/main/java/org/apache/commons/rdf/api/RDFSyntax.java
stain: Added RDFSyntax.iri() in 81d5b32 with a simplified private `IRI` implementation. I also changed the `.name()` of w3c syntaxes to match the suffix of the format URI.

(This means there are currently two entries for `RDFSyntax.name()=="RDFa"`)

stain:
Let's discuss the mutability etc. of the `RDFParser` factory separate on dev@commons - that's a general thing which anyway should not affect this pull request on `RDFSyntax`.

This pull requests tries to address [COMMONSRDF-47](https://issues.apache.org/jira/browse/COMMONSRDF-47) which rightfully complains about `RDFSyntax` being an enum instead of interface.

@ansell - would you change your `-1` (remember any committer has veto, and any ASF committer is Commons committer :-) ) if I add the plural `RDFSyntax.mediaTypes()` and `RDFSyntax.fileExtensions()` ?

ansell:
Service discovery for other instances of RDFSyntax/Parsers can be done separately. 

The main objection was for users of the RDFSyntax interface to be able to offer multiple content types and file extensions in addition to the official versions. They can be implemented by a default method that just returns the official content type and file extension inside of Commons RDF, but the fact that the method is available makes it possible to support the use case. 

Returning a List, with the first element by convention being the official version similar to RDF4J may be useful, rather than an unordered Set. However, that isn't necessary, given they should be using the other methods if they definitely want the official extension/content type.

stain:
OK, so I added them as `Set` - it might be an ordered set (e.g. `LinkedHashSet`).  

I merged `RDFA_XHTML` and `RDFA_HTML` and made them return such a set to include both `text/html` and `application/xhtml`.   Should we redefine `.equals()` and `hashCode()` to use `uri()` or would that be tricky then for arbitrary extra RDFSyntax-es?  (They could use a `urn:uuid` or something if they don't know)

stain:
I've merged this. Thanks for the help, @ansell!

PR #28, opened 2017-01-03
COMMONSRDF-52 set distinct Bundle-SymbolicName values

acoburn:
In the 0.3.0-incubating release, all of the commons-rdf components have the same `Symbolic-BundleName`, namely: `org.apache.commons.rdf`. This makes it impossible to install multiple bundles in an OSGi container (e.g. Karaf), as every bundle must have a distinct Symbolic-BundleName/Version combination.

The commons-parent pom configuration provides a mechanism for setting the value of `Symbolic-BundleName` in the bundle manifest, as shown in this PR.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -33,6 +33,10 @@
     <name>Commons RDF API</name>
     <description>Commons Java API for RDF 1.1</description>
 
+    <properties>
+      <commons.osgi.symbolicName>org.apache.commons.rdf.api</commons.osgi.symbolicName>
+    </properties>
+
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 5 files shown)


stain:
Thanks! Merged. I also added you as a `<contributor>` to be shown on https://commons.apache.org/proper/commons-rdf/team-list.html - hope that's OK.

Any other OSGi issues? How do you do the equivalent of `ServiceLoader` of `RDF` instances? Do we need to make OSGi service annotations/XML?

acoburn:
@stain for a `ServiceLoader` equivalent in OSGi, it would make sense to add a `Require-Capability` and `Provide-Capability` to each implementation's `MANIFEST.MF` -- it will simply be some minor maven changes, and I can add that in a forthcoming pull request.

PR #31, opened 2017-01-25
COMMONSRDF-54: overloads of RDF4J#asRDFTerm

ghxiao:
Resolves: COMMONSRDF-54

This PR adds three overloaded methods into the class RDF4J:

RDF4JBlankNode asRDFTerm(final org.eclipse.rdf4j.model.BNode value)
RDF4JLiteral asRDFTerm(final org.eclipse.rdf4j.model.Literal value)
RDF4JIRI asRDFTerm(final org.eclipse.rdf4j.model.IRI value)


stain:
Thanks, this looks sensible to me! 

Should we add unit test for each of the new `asRDFTerm` methods? They are in a way already tested through `createLiteral` etc.., but just so we don't break it later?

ghxiao:
I added  a test case`RDF4JMethodOverloadsTest` and one more method in RDF4J

`public RDF4JBlankNodeOrIRI asRDFTerm(final org.eclipse.rdf4j.model.Resource value)`

PR #33, opened 2017-03-10
COMMONSRDF-59 - fix minor javadocs warnings

acoburn:
This add some missing `@return` values, fixes an HTML error and adds some text to two otherwise undocumented parameters.


wikier:
LGTM

PR #34, opened 2017-03-10
COMMONSRDF-60: update commons-parent version

acoburn:
This upgrades the parent pom to the latest version of `commons-parent`.


wikier:
LGTM

PR #38, opened 2017-09-12
COMMONSRDF-62 japicmp-maven-plugin breaking build because there is atleast one incompatibility

lewismc:
This issue addresses https://issues.apache.org/jira/browse/COMMONSRDF-62

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,9 @@ Building has been tested with [Apache Maven 3.2](http://maven.apache.org/downloa
     [INFO] Final Memory: 22M/309M
     [INFO] ------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +102,9 @@ Building has been tested with [Apache Maven 3.2](http://maven.apache.org/downloa
     [INFO] Final Memory: 22M/309M
     [INFO] ------------------------------------------------------------------------
 
+If you experience issues building on Mac OSX, please use the following command
```
ansell: This isn't unique to Mac OSX, it happens for me on Linux, and I imagine it would happen on other platforms also.

PR #39, opened 2017-09-13
COMMONSRDF-62 : Ignore japicmp by default while version is 0.x

ansell:
japicmp is activated by commons-parent profile with it set to break the build on any incompatibilities. In order to have it still run by default, but not break the build, an activeByDefault=true profile is added, which will be switched off by activating any other profiles explicitly, including the release profile, but which allows development of 0.x without interruption from japicmp.

This is either an alternative, or in addition to, https://github.com/apache/commons-rdf/pull/38

The core reason is that japicmp is not recognising 0.3 -> 0.4 as a major version jump, in Semantic Versioning terms. It is treating that as a minor version jump, where the API incompatibilities still apply.

Signed-off-by: Peter Ansell <p_ansell@yahoo.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -450,22 +450,21 @@
                         </lifecycleMappingMetadata>
                     </configuration>
                 </plugin>
-          <plugin>
-        <!-- Check if we broke compatibibility against previous release -->
-        <groupId>com.github.siom79.japicmp</groupId>
-        <artifactId>japicmp-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


lewismc:
+1 @ansell this fixes 62 and should therefore supersede #38

acoburn:
👍  this is great. Thanks!

wikier:
LGTM

PR #41, opened 2017-09-28
COMMONSRDF-65

ajs6f:
https://issues.apache.org/jira/browse/COMMONSRDF-65

Also a small fix for problems building on Mac, and includes https://github.com/apache/commons-rdf/pull/39/ to fix COMMONSRDF-62 (just so that I could build! :) )


stain:
Thanks! 

This PR mixed up a couple of things which made it a bit tricky to merge - sorry for not merging #39 earlier!

2982446 merged fine and fixes [COMMONSRDF-65](https://issues.apache.org/jira/browse/COMMONSRDF-65) -- cheers!  (I'll update RDF4j to 2.2.2 as well)

abdb6ca does indeed fix [COMMONSRDF-63](https://issues.apache.org/jira/browse/COMMONSRDF-63) on my OS X machine, but I get now build errors on Windows in your new `compareBases`:

```
Tests run: 11, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.28 sec <<< FAILURE! - in org.apache.commons.rdf.simple.experimental.AbstractRDFParserTest
parseFile(org.apache.commons.rdf.simple.experimental.AbstractRDFParserTest)  Time elapsed: 0.079 sec  <<< ERROR!
java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:///C:/Users/stain/AppData/Local/Temp/test2836546964601538785.nt
        at org.apache.commons.rdf.simple.experimental.AbstractRDFParserTest.compareBases(AbstractRDFParserTest.java:139)
        at org.apache.commons.rdf.simple.experimental.AbstractRDFParserTest.parseFile(AbstractRDFParserTest.java:122)
```

I think this was caused by a dangerous pattern I have spotted with other colleagues - using `java.nio.file.Paths` to parse relative URI path strings:

```
Path basePath = Paths.get(baseUri.substring(1, baseUri.length() - 1));
```

This might work well enough on OSX and Linux, but on Windows `:` is a special character (`C:`) and the path separator is `\` rather than `/`.  `Paths` use the `FileSystems.getDefault()`.


Instead I fixed this with a simple:

`assertEquals("<" + testNt.toRealPath().toUri() + ">", firstPredicate(g, "base"));`

and then we can avoid the suggested new `parseFileSymlink()`.

(See [COMMONSRDF-63](https://issues.apache.org/jira/browse/COMMONSRDF-63))

ajs6f:
No, there were no relative URIs being parsed into `Path`s in my code-- that string munging was to remove the `<` and `>` around the URI as retrieved from the RDF. But I'm a-okay with whatever gets a `toRealPath()` in there (because of the Mac symlinked `/tmp`) and avoids any confusion over leading slashes. I can see that I created more problems with Windows.

Sorry for mixing things up-- please always feel free to kick stuff back if you feel it's liable to mess up the history or confuse people-- I can always redo the PR!

All speed to 1.0!

PR #42, opened 2017-10-30
COMMONSRDF-66:  fixes RIOT exception in JenaDatasetImpl 

christopher-johnson:
Lang.NT is not supported by registryDataset.  Opts for Lang.NQUADS as a default replacement.
adds test

### DIFF
--- a/commons-rdf-jena/src/main/java/org/apache/commons/rdf/jena/impl/JenaDatasetImpl.java
+++ b/commons-rdf-jena/src/main/java/org/apache/commons/rdf/jena/impl/JenaDatasetImpl.java
@@ -149,7 +149,7 @@ class JenaDatasetImpl implements JenaDataset {
     @Override
     public String toString() {
         final StringWriter sw = new StringWriter();
-        RDFDataMgr.write(sw, datasetGraph, Lang.NT);
+        RDFDataMgr.write(sw, datasetGraph, Lang.NQUADS);
         return sw.toString();
     }
 
--- a/commons-rdf-jena/src/test/java/org/apache/commons/rdf/jena/DatasetJenaTest.java
+++ b/commons-rdf-jena/src/test/java/org/apache/commons/rdf/jena/DatasetJenaTest.java
@@ -15,11 +15,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.commons.rdf.jena;
 
+import static org.junit.Assert.assertEquals;
+
 import org.apache.commons.rdf.api.AbstractDatasetTest;
+import org.apache.commons.rdf.api.IRI;
+import org.apache.commons.rdf.api.Literal;
 import org.apache.commons.rdf.api.RDF;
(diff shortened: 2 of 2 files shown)


jena/src/test/java/org/apache/commons/rdf/jena/DatasetJenaTest.java
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
```
afs: Accidental reformat of the license.

jena/src/main/java/org/apache/commons/rdf/jena/impl/JenaDatasetImpl.java
```diff
@@ -149,7 +149,7 @@ public long size() {
     @Override
     public String toString() {
         final StringWriter sw = new StringWriter();
-        RDFDataMgr.write(sw, graph, Lang.NT);
```
afs: It would good to rename graph (throughout the class) as `datasetGraph` or some such. It's not a graph.

christopher-johnson: right.  would something like this be ok?  
```java
    private final DatasetGraph dg;

    JenaDatasetImpl(final DatasetGraph datasetGraph, final UUID salt) {
        this.dg = datasetGraph;
        this.salt = salt;
        this.factory = new JenaRDF(salt);
    }
```
so `dg` is the private variable, and `datasetGraph` is only used in the constructor to match InternalJenaFactory.

jena/src/test/java/org/apache/commons/rdf/jena/DatasetJenaTest.java
wikier: Please, remove this changeset from the PR.

jena/src/main/java/org/apache/commons/rdf/jena/impl/JenaDatasetImpl.java
wikier: I'd keep both, parameter and attribute, as `datasetGraph`.

wikier:
Please, rebase this PR on top of the current `HEAD` of `master`.

jena/src/main/java/org/apache/commons/rdf/jena/impl/JenaDatasetImpl.java
wikier: Attribute already renamed in `master` (see 151a8ea). 

So, please, rebase this PR to the current `HEAD` so we can ship this patch in the imminent `0.5.0` release.

christopher-johnson:
CI error is currently noted by Travis as an [open issue](https://www.traviscistatus.com/incidents/xgpn2tq6mx7k).

wikier:
LGTM

PR #44, opened 2017-11-03
COMMONSRDF-70: Upgrade Jena version to 3.5.0

ajs6f:


ajs6f:
Definitely can wait. There's no substantive API changes at stake, and anyone who wants to use Jena 3.5.0 w/ Commons RDF should be able to override the transitive dependency without untoward effect.

acoburn:
In fact, I can confirm that commons-rdf-jena/0.5.0 works fine with jena/3.5.0

wikier:
Now I think this should definitively go in `0.5.0-RC2`.

ajs6f:
👍

PR #45, opened 2017-11-20
COMMONSRDF-71: Update Javadocs link for RDF4J

acoburn:
Updates location of RDF4J javadocs link


wikier:
Good catch. Maybe at some point we just want to have http://docs.rdf4j.org/javadoc/latest/ but LGTM.

PR #49, opened 2017-12-14
Cleanup for FindBugs and PMD warnings in -simple and -jena

ajs6f:


commons-rdf-simple/src/main/java/org/apache/commons/rdf/simple/experimental/AbstractRDFParser.java
```diff
@@ -58,8 +60,12 @@
  */
 public abstract class AbstractRDFParser<T extends AbstractRDFParser<T>> implements RDFParser, Cloneable {
 
-    public static final ThreadGroup threadGroup = new ThreadGroup("Commons RDF parsers");
-    private static final ExecutorService threadpool = Executors.newCachedThreadPool(r -> new Thread(threadGroup, r));
+    public static final AtomicInteger threadCount = new AtomicInteger();
+    private static Thread newThread(Runnable r) {
```
stain: I disagree on this style change, the previous code was much cleaner. Why is a method-reference on 4 lines that is only used once better than a very small lambda?

ajs6f: Because either FindBugs or PMD (can't remember) which called out `ThreadGroup` as not threadsafe, which is correct as far as I know. I am in _no_ way wedded to this change and if we can guarantee thread-safety from outside the class (or simply document it as not threadsafe) I would be happy to pull this change out.

stain: A final ThreadGroup that is not thread safe.. The whole purpose of it is to group threads?

It is true that they should not be used for security purpose, but that is not why it is here. The ThreadGroup is mainly useful for debugging as people might see these "Commons RDF Parser" tree grouped together in the debugger (e.g. in Eclipse) rather than the generic no-name you get from the ExecutorService.

ajs6f: That's why I put [the name "Commons RDF Parser "](https://github.com/apache/commons-rdf/pull/49/files/f1649e034a2623137434fcc2810f8802d3ee9434#diff-68c2acaf43f1ae22da0bdb4eac55a201R65) back in. But like I said, I am fine with whatever- I was just tearing through PMD warnings at speed after the comment (I think to the board report) about there being a lot of them. Shall I take your comment as a direction to remove this change? Happy to...

PR #53, opened 2018-12-29
COMMONSRDF-80: Bumping Jena version 3.5.0 => 3.9.0

ajs6f:


acoburn:
The jena dependency is currently tied to the `jena-osgi` artifact. I think it would be much better to set the dependency to `jena-arq`. This way, in a non-OSGi context, dependency resolution is easier to manage; in an OSGi context, a developer would still use the `jena-osgi` module at deploy time.

ajs6f:
As another Jena committer, I'd be happier with that too. Thoughts? @stain ? @wikier? Others?

stain:
It is not as ideal, as it would make `commons-rdf-jena` as a Maven package no longer be a sufficient OSGi dependency by itself, as `jena-arq` is not an OSGi package (for some reason). 

But I understand it would make it easier for anyone using the regular `jena-arq`  dependencies together with Commons RDF to avoid duplication (particularly if you get different versions of Jena).

I think for the OSGi dependencies to be calculated correct for `commons-rdf-jena` we need to then still list `jena-osgi` as an `<optional>` dependency.

While the ideal solution would be for `jena-arq` etc. to be OSGi packages themselves (avoiding need for `jena-osgi` overall) I'm OK with the above solution for `commons-rdf-jena`. OSGi users would simply need to ignore `jena-arq` and add `jena-osgi`, we can document that.

stain:
OK, changed to `apache-jena-libs` in ea52c2ff, also [documented jena-osgi use](https://github.com/ajs6f/commons-rdf/blob/COMMONSRDF-80/src/site/markdown/implementations.md#jena-and-osgi)

It seems `apache-jena-libs` also adds in `jena-tdb2` and legacy `jena-tdb` - perhaps this is a bit excessive? Could you verify @ajs6f - it's what is documented on https://jena.apache.org/download/maven.html so presumably this would help the transitive `jena-*` all be in the same version, or would we need a `<dependencyManagement>`?

stain:
Great, would you be able to update `implementations.md` on how to use the `jena-osgi-features` as an alternative? I'll change to `jena-arq` which should be sufficient.

wikier:
LGTM

PR #205, opened 2024-03-09
GraphLike with bulk add

naturzukunft:
Snippet from dev@commons.apache.org:
```
Hi there,
it looks like I am bound to the interface org.apache.commons.rdf.api.GraphLike#add. And in the RDF4J implementation, a new connection to the repository is established for each quad/triple. This is not very useful.
Am I missing another possibility or am I the first to have a problem with this ?

Is there a reason, why there is no Method like GraphLike#add(List<T> statements); ?

Thanks Fredy_
```
PR extends the Interface `GraphLike` and do optimized bulk add to rdf4j Repository in one connection.


naturzukunft:
Hi there, is there a problem with this PR?
Why is nobody approve the build ?
Thanks

garydgregory:
We're busy obviously 😉 I'll take a look over the next few days.

pedro-w:
Just to keep everything in one place I made some comments [here](https://lists.apache.org/thread/m020v4wr8fcwjvr417gb3fykdzlt772b) which I have reproduced below

1. Better IMO to add a default method to the interface rather than adding 11 implementations which are almost identical
2. No need to restrict it to a List when any iterable would do
3. There is a bulk 'add' method in RepositoryConnection (https://rdf4j.org/javadoc/latest/org/eclipse/rdf4j/repository/RepositoryConnection.html#add(java.lang.Iterable,org.eclipse.rdf4j.model.Resource...) - might be better to use it
4. Why was RDF4JServiceLoaderTest.java deleted?
5. Method name 'addAll' is more consistent with Java collections but it seems RDF4J uses 'add' overloaded for single or multiple additions - don't know if Commons has any preference on that.

and also in [that thread](https://lists.apache.org/thread/rs4d5tzpf3sx0fxllhx3yy8gbczvksnx) I added a patch which is related to 1 & 2 on the list above.

PR #324, opened 2025-04-17
upgrade rdf4j to 5.1.3 (Tue, Apr 15, 2025)

naturzukunft:
With this PR commons-rdf is using the newest version (Tue, Apr 15, 2025) of rdf4j.

### DIFF
--- a/commons-rdf-integration-tests/src/test/resources/alice-cached.jsonld
+++ b/commons-rdf-integration-tests/src/test/resources/alice-cached.jsonld
@@ -1,8 +1,8 @@
 { "http://purl.org/dc/terms/license" :
    "Licensed to the Apache Software Foundation (ASF) under one or more\n contributor license agreements.  See the NOTICE file distributed with\n this work for additional information regarding copyright ownership.\n The ASF licenses this file to You under the Apache License, Version 2.0\n (the \"License\"); you may not use this file except in compliance with\n the License.  You may obtain a copy of the License at\n \n http://www.apache.org/licenses/LICENSE-2.0\n \n Unless required by applicable law or agreed to in writing, software\n distributed under the License is distributed on an \"AS IS\" BASIS,\n WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n See the License for the specific language governing permissions and\n limitations under the License.\n",
   
-  "@context": "http://example.com/context",
-  "@id": "ex:Alice",
+  "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+  "@id": "http://example.com/Alice",
   "@type": "Person",
   "name": "Alice W. Land"
 }
--- a/commons-rdf-rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4J.java
+++ b/commons-rdf-rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/RDF4J.java
@@ -109,7 +109,7 @@ public final class RDF4J implements RDF {
          */
         includeInferred,
         /**
-         * The graph/dataset should handle {@link Repository#initialize()} (if
+         * The graph/dataset should handle {@link Repository#init()} (if
          * needed) and {@link Repository#shutDown()} on {@link Graph#close()} /
          * {@link Dataset#close()}.
          */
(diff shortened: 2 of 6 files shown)


garydgregory:
Hi @naturzukunft 

Would you please look into fixing the Java 21 build? I'm not comfortable with changes until we get builds on Java LTS versions fixed.

TY!

naturzukunft:
> Hi @naturzukunft
> 
> Would you please look into fixing the Java 21 build? I'm not comfortable with changes until we get builds on Java LTS versions fixed.
> 
> TY!

I think it needs a jena expert to upgrade to a newer jena version! commons-rdf is using jena 3.5.0 from Oct 30, 2017.
There are dependencies to artifacts, that are not available in the current jena release. I think i'm not the right one to do this upgrade.

garydgregory:
> > Hi @naturzukunft
> > Would you please look into fixing the Java 21 build? I'm not comfortable with changes until we get builds on Java LTS versions fixed.
> > TY!
> 
> I think it needs a jena expert to upgrade to a newer jena version! commons-rdf is using jena 3.5.0 from Oct 30, 2017. There are dependencies to artifacts, that are not available in the current jena release. I think i'm not the right one to do this upgrade.

Hello @naturzukunft 

Do you think upgrading Jena is _required_ to run this component on Java 21? That would be unusual.

naturzukunft:
> > > Hi @naturzukunft
> > > Would you please look into fixing the Java 21 build? I'm not comfortable with changes until we get builds on Java LTS versions fixed.
> > > TY!
> > 
> > 
> > I think it needs a jena expert to upgrade to a newer jena version! commons-rdf is using jena 3.5.0 from Oct 30, 2017. There are dependencies to artifacts, that are not available in the current jena release. I think i'm not the right one to do this upgrade.
> 
> Hello @naturzukunft
> 
> Do you think upgrading Jena is _required_ to run this component on Java 21? That would be unusual.

at the moment i have no idea what the cause of the error is.

naturzukunft:
The original `org.apache.commons.rdf.api.AbstractDatasetTest`:

```
    @Test
    public void testStreamDefaultGraphNameByPattern() throws Exception {
        // Explicitly select in only the default graph Optional.empty()
        final Optional<? extends Quad> aliceTopic = dataset.stream(Optional.empty(), null, null, null).findAny();
        assertTrue(aliceTopic.isPresent());
        // COMMONSRDF-55: should not be <urn:x-arq:defaultgraph> or similar
        assertNull(aliceTopic.get().getGraphName().orElse(null));
        assertFalse(aliceTopic.get().getGraphName().isPresent());
    }
```
is running with jdk 17.

if i change it to this, which I think tests the same thing only slightly better, then the test also fails with jdk 17:
```
    @Test
    public void testStreamDefaultGraphNameByPattern() throws Exception {
        dataset.stream(Optional.empty(), null, null, null).forEach(quad -> {
            // COMMONSRDF-55: should not be <urn:x-arq:defaultgraph> or similar
            assertNull(quad.getGraphName().orElse(null));
            assertFalse(quad.getGraphName().isPresent());
        });
    }
```

So in my opinion the test is wrong. And the order of the stream changes with a newer jdk, so `findAny()`  delivers another quad.

naturzukunft:
See: https://github.com/apache/commons-rdf/pull/327

garydgregory:
Hi @naturzukunft 

Please rebase on git master.

TY!

naturzukunft:
@garydgregory is there a snapshot repo, where this version is available ?

naturzukunft:
> @garydgregory is there a snapshot repo, where this version is available ?

Found it [here](https://commons.apache.org/proper/commons-rdf/userguide.html#Using_Commons_RDF_from_Maven). Was expecting it in [Downloads](https://commons.apache.org/proper/commons-rdf/download.html).

The url is wrong, it has to be `https`

naturzukunft:
But my pipeline is downloading an old version ;-(

Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/commons/commons-rdf-api/0.6.0-SNAPSHOT/commons-rdf-api-0.6.0-20250322.134200-21.pom

```xml
<repository>
	<id>apache.snapshots</id>
	<name>Apache Snapshot Repository</name>
	<url>https://repository.apache.org/snapshots</url>
	<releases>
		<enabled>false</enabled>
	</releases>
</repository>
```
expected: https://repository.apache.org/content/groups/snapshots/org/apache/commons/commons-rdf-api/0.6.0-SNAPSHOT/commons-rdf-api-0.6.0-20250124.151445-20.jar

i changed now the url to:

```xml
<repository>
	<id>apache.snapshots</id>
	<name>Apache Snapshot Repository</name>
	<url>https://repository.apache.org/content/groups/snapshots</url>
	<releases>
		<enabled>false</enabled>
	</releases>
</repository>
```

also the newest version is not loaded.

In [maven-metadata.xml](https://repository.apache.org/content/groups/snapshots/org/apache/commons/commons-rdf-api/0.6.0-SNAPSHOT/maven-metadata.xml) the latest version is `<lastUpdated>20250322134200</lastUpdated>`, what feels somehow wrong?

Any idea ?

PR #328, opened 2025-04-23
stream() implementations should default to being sequential

naturzukunft:
as mentioned in the mailing list:

```
Subject: [RDF] ModelGraphImpl returns 'parallelStream()' instead of expected 'stream()'
Date: Thu, 17 Apr 2025 18:18:15 +0200
From: Fred Hauschel <info@hauschel.de>
To: dev <dev@commons.apache.org>

Hi there, in my opinion this is a bug:

 * https://github.com/apache/commons-rdf/blob/006fbc3e47003c2d0fe8eb055f60f7cb267e1321/commons-rdf-rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java#L142
 * https://github.com/apache/commons-rdf/blob/006fbc3e47003c2d0fe8eb055f60f7cb267e1321/commons-rdf-rdf4j/src/main/java/org/apache/commons/rdf/rdf4j/impl/ModelGraphImpl.java#L150

Because nobody expects a parallel stream to be created and returned when stream() is called.
There is also a great danger that this parallel stream will be used to perform write operations with RDF4J. However, RDF4J is not thread-safe and this can lead to non-deterministic results.

What is your opinion on this?

Thanks Fredy
```

This PR fixes the problem in the rdf4j implementation and is backward compatible.
(in the worst case scenario, existing users will receive a stream instead of a parallel stream if they do not migrate.)

### DIFF
--- a/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
+++ b/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
@@ -321,5 +321,17 @@ public interface Graph extends AutoCloseable, GraphLike<Triple> {
      * @since 0.3.0-incubating
      */
     Stream<? extends Triple> stream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object);
-
+    
+    /**
+	 * Same as {@link #stream()}, but returns a possibly parallel {@code Stream} of
+	 * {@link TripleLike} statements.
+	 * 
+	 * @param subject   The triple subject (null is a wildcard)
+	 * @param predicate The triple predicate (null is a wildcard)
+	 * @param object    The triple object (null is a wildcard)
+	 * @return A {@link Stream} over the matched triples.
+	 */
+    default Stream<? extends Triple> parallelStream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
+    	return stream(subject, predicate, object);
+    }
 }
--- a/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
+++ b/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
@@ -323,16 +323,16 @@ public interface Graph extends AutoCloseable, GraphLike<Triple> {
     Stream<? extends Triple> stream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object);
     
     /**
-	 * Same as {@link #stream()}, but returns a possibly parallel {@code Stream} of
-	 * {@link TripleLike} statements.
-	 * 
-	 * @param subject   The triple subject (null is a wildcard)
-	 * @param predicate The triple predicate (null is a wildcard)
-	 * @param object    The triple object (null is a wildcard)
-	 * @return A {@link Stream} over the matched triples.
-	 * @since 0.6.0
-	 */
+     * Same as {@link #stream()}, but returns a possibly parallel {@code Stream} of
+     * {@link TripleLike} statements.
+     * 
+     * @param subject   The triple subject (null is a wildcard)
+     * @param predicate The triple predicate (null is a wildcard)
+     * @param object    The triple object (null is a wildcard)
+     * @return A {@link Stream} over the matched triples.
+     * @since 0.6.0
+     */
     default Stream<? extends Triple> parallelStream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
-    	return stream(subject, predicate, object);
+        return stream(subject, predicate, object);
     }
 }
--- a/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
+++ b/commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
@@ -330,6 +330,7 @@ public interface Graph extends AutoCloseable, GraphLike<Triple> {
 	 * @param predicate The triple predicate (null is a wildcard)
 	 * @param object    The triple object (null is a wildcard)
 	 * @return A {@link Stream} over the matched triples.
+	 * @since 0.6.0
 	 */
     default Stream<? extends Triple> parallelStream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object) {
     	return stream(subject, predicate, object);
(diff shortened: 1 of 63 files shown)


naturzukunft:
> -1 since there are no unit tests.

In my opinion, the changes are 100% delegation to rdf4j classes. There is no new logic that can be tested. If I add tests, they would only test org.eclipse.rdf4j.*. Can you specify which logic should be tested?

naturzukunft:
@since 0.6.0 added

naturzukunft:
@garydgregory please do not merge this PR yet. i was able to run my tests with 0.6.0-SNAPSHOT and have a strange behaviour. i have to find that problems first.

naturzukunft:
I am 99% sure that my problems are related to the use of `parallelStream` in `ModelGraphImpl`. when I run my test in the local ide, I get a different error every time I run it. This points to a concurrency problem. if i replace `parallelStream` with `stream` the problems are gone. So i expect merging this PR would solve my problems.

garydgregory:
Hello @naturzukunft 

Thank you for your updates. I just did a bit of housekeeping. Would you mind rebasing this PR on git master?

naturzukunft:
> Hello @naturzukunft
> 
> Thank you for your updates. I just did a bit of housekeeping. Would you mind rebasing this PR on git master?

done

naturzukunft:
@garydgregory i created a new test branch from synced master, the pipeline also fails. So it looks like a common problem.
Can you please approve the pipelines and let me know what stands in the way of a merge

garydgregory:
I'll review the PRs this week.

naturzukunft:
> I'll review the PRs this week.

@garydgregory Friendly reminder :-)

garydgregory:
Hello @naturzukunft 

I've taken a look again and here's what I propose:
- Since this is not 1.0 yet, we can change the behavior and APIs, which here feels like a bug anyway.
- Since `GraphLike` is not a `Collection`, it does not need to define a `parallelStream()` API.
- Change `stream()` implementations to return a plain (IOW sequential) stream.
- If a call site wants a parallel stream, it should call `stream().parallel()` per:
  - `java.util.stream.BaseStream.parallel()`
  - `java.util.stream.BaseStream.sequential()`
- The above avoids increasing the public API surface with seemingly redundant APIs.

All the above should be OK unless there's something in the domain model that requires implementations to be parallel, so you see such a requirement?

naturzukunft:
> Hello @naturzukunft
> 
> I've taken a look again and here's what I propose:
> 
>     * Since this is not 1.0 yet, we can change the behavior and APIs, which here feels like a bug anyway.
> 
>     * Since `GraphLike` is not a `Collection`, it does not need to define a `parallelStream()` API.
> 
>     * Change `stream()` implementations to return a plain (IOW sequential) stream.
> 
>     * If a call site wants a parallel stream, it should call `stream().parallel()` per:
>       
>       * `java.util.stream.BaseStream.parallel()`
>       * `java.util.stream.BaseStream.sequential()`
> 
>     * The above avoids increasing the public API surface with seemingly redundant APIs.
> 
> 
> All the above should be OK unless there's something in the domain model that requires implementations to be parallel, so you see such a requirement?

commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
```diff
@@ -321,5 +321,4 @@ default Iterable<Triple> iterate(final BlankNodeOrIRI subject, final IRI predica
      * @since 0.3.0-incubating
      */
     Stream<? extends Triple> stream(BlankNodeOrIRI subject, IRI predicate, RDFTerm object);
-
```
garydgregory, line 324: Please remove style change from the PR.

commons-rdf-api/src/main/java/org/apache/commons/rdf/api/GraphLike.java
```diff
@@ -102,5 +102,4 @@ public interface GraphLike<T extends TripleLike> {
      * @return A {@link Stream} of {@link TripleLike} statements.
      */
     Stream<? extends T> stream();
-
```
garydgregory, line 105: Please remove style change from the PR.

commons-rdf-api/src/test/java/org/apache/commons/rdf/api/DummyGraphTest.java
```diff
@@ -83,5 +83,4 @@ public void testStreamFiltered() throws Exception {
                 graph.stream(new DummyIRI(1), new DummyIRI(2), new DummyIRI(3)).findAny().get());
         assertFalse(graph.stream(new DummyIRI(0), new DummyIRI(0), new DummyIRI(0)).findAny().isPresent());
     }
-
```
garydgregory, line 86: Please remove style change from the PR.

naturzukunft:
Hi @garydgregory,
that works for me. I have removed the parallelStream methods again.
If there is anything else to do, please let me know.
Fredy

garydgregory (changes requested):
Hello @naturzukunft 

Thank you for updating the PR.
Please remove style changes from the PR.
To avoid regressions, please add tests that check that we are indeed not returning parallel streams.

naturzukunft:
> please add tests that check that we are indeed not returning parallel streams.

@garydgregory  what do you mean? only for rdf4j ? because other graph implementations still use parallel streams. That would be a major change that I don't really want to tackle.

commons-rdf-api/src/main/java/org/apache/commons/rdf/api/Graph.java
naturzukunft: done

commons-rdf-api/src/test/java/org/apache/commons/rdf/api/DummyGraphTest.java
naturzukunft: done

naturzukunft:
Hi @garydgregory 
as an impatient person, I'm slowly struggling with my motivation ;-)
"please add tests that check that we are indeed not returning parallel streams." is now done.
Fredy

garydgregory:
Hi @naturzukunft 

I fixed some issues in the build. Would you please rebase on git master?

naturzukunft:
> Hi @naturzukunft
> 
> I fixed some issues in the build. Would you please rebase on git master?

done

garydgregory:
Hello @naturzukunft 
One more rebase, please. I had to make some more changes to get reliable builds.
TY.

naturzukunft:
@garydgregory done.

garydgregory:
@naturzukunft 
Merged, ty.

